PR #12, opened 2015-07-23
Change default argument behavior for getting an EM with no arguments to manager

FoxxMD:
Allows more options for getting the default EM by looking first for any connections or managers with the name 'default', then if none exists fallback to head


FoxxMD:
@mbbender @patrickbrouwers what do you think?

FoxxMD:
Not a big deal I think.

PR #15, opened 2015-07-27
Do not defer provider

FoxxMD:
Do not defer provider during application boot and extend auth manager during registration. This fixes a bug where the auth manager cannot access entities because laraveldoctrine hasn't initialized yet.

I ran in to a problem where [AuthManager](https://github.com/illuminate/auth/blob/master/AuthManager.php#L16) attempted to create a driver for a callback to view entities and threw an exception because doctrine hadn't been initialized. 

The solution I found to fix this was to call `extendAuthManager()` in `register` and change the provider to not be deferred. This way doctrine was ready to go by the time the auth manager needs to make a call.

### DIFF
--- a/src/DoctrineServiceProvider.php
+++ b/src/DoctrineServiceProvider.php
@@ -49,7 +49,7 @@ class DoctrineServiceProvider extends ServiceProvider
      * Indicates if loading of the provider is deferred.
      * @var bool
      */
-    protected $defer = true;
+    protected $defer = false;
 
     /**
      * Boot service provider.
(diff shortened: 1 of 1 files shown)


FoxxMD:
@mbbender @patrickbrouwers What do you think?

patrickbrouwers:
Shouldn't be necessary. As soon as the AuthManager class get's called the DoctrineSP will be booted (provides array). 

What's the use-case this goes wrong? I didn't run into any problems.

Perhaps the extendAuthManager can be moved to register without changing the `defer` setting. The L5-docs aren't really clear about it: http://laravel.com/docs/5.0/extending#authentication Session and Cache must be in boot, but probably Auth is possible in register.

mbbender:
Hum, I would agree with Patrick that we should not have to change defer to true. I'm not sure how the extend() function works to know if register or boot makes sense, but from what I can tell we should only be telling Laravel what concrete implementations to use for injected interfaces in the register method. We shouldn't be doing anything else in there.

I also think we need to remove extendAuth from the doctrine service provider altogether and make it a DoctrineAuthServiceProvider class.

PR #17, opened 2015-07-27
Create instance of subscriber

mbbender:
Seems the addEventSubscriber function is expecting an EventSubscriber, not a string to a class. I beleive we should be creating the subscriber before we add it. Hows it look @patrickbrouwers @FoxxMD ?


patrickbrouwers:
:+1:

PR #22, opened 2015-08-05
Subscribe extensions only once.

jee7:
Fix for ExtensionManager subscribing the same extensions on the same entity manager multiple times.
ExtensionManager is a singleton, but the DoctrineServiceProvider is not. Furthermore, DoctrineServiceProvider is configured to be a deferred service for several things (entity manager, auth manager etc). This means, that the DoctrineServiceProvider is instantiated several times. It calls boot() on the ExtensionManager singleton several times. Alas the different extensions should subscribe to a given entity manager only once.
I think it's the ExtensionManager's job to keep track of extensions it subscribes to an entity manager.

### DIFF
--- a/src/Extensions/ExtensionManager.php
+++ b/src/Extensions/ExtensionManager.php
@@ -40,6 +40,11 @@ class ExtensionManager
     protected $registry;
 
     /**
+     * @var array
+     */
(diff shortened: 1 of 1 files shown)


patrickbrouwers:
+1

PR #25, opened 2015-08-06
Call a register() method if an extension provides one.

jee7:
This is required for the UploadableExtension that registers the listener as a singleton (in order to access it easily from the app).
Perhaps it would be better to do this using the interface (although most of the extensions do not register anything) or hierarchy (create an AbstractExtension that provides an empty register() method). I'm not sure if calling method_exists() is slower or faster than calling an empty method.


jee7:
No need to do that if the Uploadable extension has it's own service provider.

PR #26, opened 2015-08-13
Fix Composer.json

kayladnls:
@rdohms will get you.


patrickbrouwers:
@rdohms give us a 5 minute head start ;)

PR #28, opened 2015-08-29
Easy way to hook into Doctrine configuration

patrickbrouwers:
Suggestion for an easy way to hook into doctrine's configuration, connection and event manager, without needing to have an extensive knowledge of doctrine and without us having to specify a lot of config options which most people don't use by default.

The DoctrineManager class has a `extend` method which excepts the manager name (connection) from the managers array in the config. The second argument is the callable (which can be class or closure) and returns instances of `Configuration`, `Connection` and `EventManager` for the given connection.
Classes should implement `LaravelDoctrine\ORM\DoctrineExtender` and provide an `extend` method.

```php
$manager = app( LaravelDoctrine\ORM\DoctrineManager::class);

$manager->extend('default', function(Configuration $configuration, Connection $connection, EventManager $eventManager) {
    // Do stuff with configuration for given connection
});

$manager->extend('default', MyDoctrineExtender::class);

class MyDoctrineExtender implements \LaravelDoctrine\ORM\DoctrineExtender
{
    /**
     * @param Configuration $configuration
     * @param Connection    $connection
     * @param EventManager  $eventManager
     */
    public function extend(Configuration $configuration, Connection $connection, EventManager $eventManager)
    {
        // Do stuff with configuration for given connection
    }
}
```

If you want to apply the settings to all connections, you can use `extendAll` which does the exact same as `extend` except it loops through all entity managers registered in the registry.

```php
$manager->extendAll(function(Configuration $configuration, Connection $connection, EventManager $eventManager) {
    // Do stuff with configuration for all connections
});
```

For ease I added a facade as well:

```php
Doctrine::extend('default', function(Configuration $configuration) {
    // Do stuff with configuration for given connection
});
```

Edit: I used this as an example for Beberlei extensions.


FoxxMD:
I like it :+1:

josenicomaia:
+1 =]

PR #46, opened 2015-09-23
Added the ability to add multiple listeners to each event.

maxbrokman:
In 1.0 release each Doctrine event can only have one listener attached to it, but it would be helpful to be able to add many listeners to a single event (the same effect can currently be achieved by adding multiple subscribers). This PR should have no BC break, it checks if the listener passed is just a single class or an array and behaves appropriately.


patrickbrouwers:
Good job @maxbrokman . Thanks for including the tests as well!

PR #49, opened 2015-09-24
Made the `Jsonable` trait compatible with Laravel's `Jsonable` interface

maxbrokman:
Added the ability to pass options to the json serializer to make the definition of the trait `Jsonable` compatible with Laravel's `Jsonable` interface.

It would be nice to be able to return Laravel style collections from repositories. These include a `toJson` method, which looks for an object that implements `Illuminate\Contracts\Support\Jsonable`. In order to fulfil this interface we need to be able to accept Json options (such as `JSON_NUMERIC_CHECK`) in our trait.

### DIFF
--- a/src/Serializers/JsonSerializer.php
+++ b/src/Serializers/JsonSerializer.php
@@ -26,11 +26,12 @@ class JsonSerializer
     /**
      * @param $entity
      *
+     * @param  int    $jsonEncodeOptions
      * @return string
      */
-    public function serialize($entity)
+    public function serialize($entity, $jsonEncodeOptions = 0)
     {
-        return $this->serializer->serialize($entity, 'json');
+        return $this->serializer->serialize($entity, 'json', ['json_encode_options' => $jsonEncodeOptions]);
     }
 
     /**
--- a/src/Serializers/Jsonable.php
+++ b/src/Serializers/Jsonable.php
@@ -5,11 +5,12 @@ namespace LaravelDoctrine\ORM\Serializers;
 trait Jsonable
 {
     /**
+     * @param  int    $options
      * @return string
      */
-    public function toJson()
+    public function toJson($options = 0)
     {
-        return (new JsonSerializer)->serialize($this);
+        return (new JsonSerializer)->serialize($this, $options);
     }
 
     /**
--- a/tests/Serializers/JsonSerializerTest.php
+++ b/tests/Serializers/JsonSerializerTest.php
@@ -15,11 +15,20 @@ class JsonSerializerTest extends PHPUnit_Framework_TestCase
         $this->serializer = new JsonSerializer;
     }
 
-    public function test_can_serialize_to_array()
+    public function test_can_serialize_to_json()
     {
-        $array = $this->serializer->serialize(new JsonableEntity);
+        $json = $this->serializer->serialize(new JsonableEntity);
 
-        $this->assertEquals('{"id":"IDVALUE","name":"NAMEVALUE"}', $array);
+        $this->assertJson($json);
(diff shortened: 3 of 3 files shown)


patrickbrouwers:
👍

Francismori7:
I still cannot make out how to use such:

In my controller:
`return $users->findAll();`

My entity implements JsonableContract and ArrayableContract from Illuminate, as well as uses the Jsonable and Arrayable from L-D. I still get an empty response, or almost:

`[
    {},
    {},
    {},
    {}
]`

Both methods work properly on a ->find() returned entity (toJson and toArray).

patrickbrouwers:
If you could add a failing test, that would be very helpful.

FoxxMD:
:+1: for this.

maxbrokman:
@Francismori7

The first thing to understand is how Laravel works out if it can output json from what you return from the controller.

This logic is contained in `Illuminate\Http\Response::shouldBeJson`. The check looks like this:

```php
protected function shouldBeJson($content)
{
    return $content instanceof Jsonable ||
           $content instanceof ArrayObject ||
           is_array($content);
}
```

The behaviour you're probably used to as a Laravel user is your calls to Eloquent coming back as items wrapped in an instance
of `Illuminate\Support\Collection` or similar. This class implements `Jsonable` and delegates its `toJson()` behaviour
to its `toArray()` method. This method converts everything to an array which can then be passed to a naive `json_encode()`,
which is what the collection is doing.

However, your `Repository::findAll()` is returning an plain array. `Response` knows this can be converted to json but
all it does to it is a simple `json_encode()` on your array of objects. Because your objects have no public properties
you get the `[{}, {}]` you are seeing.

To make this work you can do two things (or do both and then it will work all the time!):

1. Wrap your repository results in `Illuminate\Support\Collection` and let that handle the `toJson()` behaviour when
you have a collection of users. This would entail implementing `Illuminate\Contracts\Support\Arrayable` on your
entities (because that's how the collection is going to convert them to json), and then in your repository:

```php
public function findAll()
{
    return new Illuminate\Support\Collection(parent::findAll());
}
```

2. Implement `\JsonSerializable` on your Entity. This tells `json_encode()` that it can look for `User::jsonSerialize()`
for instructions on how to convert the object to something that can be json encoded. The `Jsonable` trait contains an
implementation of `jsonSerialize()` for you so adding the interface is all you would need to do. `json_encode` on a single
object seems to check for `jsonSerialize()` automatically, but it doesn't do this on an array.

Personally I'm doing both as I like having access to the helpers (map, each etc) on the collection coming back from
my repositories.

Francismori7:
Thanks a lot @maxbrokman 

Next thing this requires is support for hidden attributes - that way we don't expose let's say, passwords hash to the public with our JSON.

patrickbrouwers:
For more complex, non-standard situations I would highly recommend to use a `Transformer` object. It will give you a lot more control over the output.
Fractal (http://fractal.thephpleague.com/) makes it very easy to implement.

bskimball:
I actually use a little bit different method. I use the instance of Collection and then use Laravel's response to return the json.

in my Repo

``` php
public function getItems() {
  $qb = $this->createQueryBuilder('i');
  $qb->select(//build your fields or whatever you intend to do with querybuilder);
  $result = new \Illuminate\Support\Collection($qb->getQuery()->getArrayResult());
  return $result;
}
```

then in my controller

``` php
public function index() {
  $items = $this->itemrepo->getItems();
  return response()->json(['items' => $items], 200);
}
```

maybe this helps someone

joaoeduardo:
How to ignore a property?

maxbrokman:
Depending on what you're trying to do you need to look into more controller serialization using something like [Fractal](http://fractal.thephpleague.com/)

PR #53, opened 2015-09-29
Allow type hints/casts in User construct

thoresuenert:
to allow type casts in the __construct function of User the getEntity function has to return a new instance without the contructor to avoid type mismatch


guiwoda:
I'm not sure that's the way to go for this... I'd rather force a default constructor than reflect on the entity, it looks too hacky!

src/Auth/DoctrineUserProvider.php
```diff
@@ -128,6 +129,8 @@ protected function getRepository()
      */
     protected function getEntity()
     {
-        return new $this->entity;
```
josenicomaia: What about?
```php
return new {$this->entity}();
```

patrickbrouwers: That will give the same error when having forced parameters in your constructor.

FoxxMD:
@guiwoda Let's me honest, the majority of PHP is magic :sweat_smile:

src/Auth/DoctrineUserProvider.php
josenicomaia: Ops., I got it!

PR #56, opened 2015-10-02
Allow specifying an extension for Yaml and Xml Metadata drivers

matthew-james:
This pull request allows the user to specify the extension used by the Xml and Yaml drivers.

BC break: No.   If the extension is not specified the doctrine default extension is used, so this will not be a bc break.

Use case:  Simplifies moving to this package if you were already using Doctrine directly with a custom extension.

Affects Documentation: Yes.  A configuration option is available called 'extension'.

I tested as well as I could, but there is not a way to view the extension once it is passed to the driver, and I cannot mock the driver class.

### DIFF
--- a/src/Configuration/MetaData/Xml.php
+++ b/src/Configuration/MetaData/Xml.php
@@ -14,7 +14,8 @@ class Xml extends MetaData
     public function resolve(array $settings = [])
     {
         return new XmlDriver(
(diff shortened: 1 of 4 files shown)


patrickbrouwers:
Could you make this PR to the 1.0 branch please as it is no breaking change (see contribution guide lines). Thanks!

PR #69, opened 2015-10-13
Allow specifying the EntityManager used by EntityManagerFactory

matt-allan:
This PR adds an optional config key 'entity_manager'.  If specified, the provided class will be used instead of the default.

Use case:

Adding the 'entity_manager' option would allow the end user to override the behavior of the entity manager returned by the EntityManagerFactory.

I personally use a custom manager similar to [https://github.com/laravel-doctrine/orm/commit/bfa8e471238bddbbd603bf9c7e0d84dffd56216d](these changes on master) to specify a custom default hydrator.

A config option would allow changing the hydration mode, as well as changing any configuration options before the Entity Manager is created.

This would also allow users who do not want to use Illuminate Collections to specify the default Entity Manager.

Requires Documentation Update: Yes.

BC Break: No.  This should also merge cleanly with the changes in master, since master only changed the use statement in the EntityManagerFactory class.

### DIFF
--- a/src/EntityManagerFactory.php
+++ b/src/EntityManagerFactory.php
@@ -111,10 +111,7 @@ class EntityManagerFactory
             array_get($settings, 'repository', EntityRepository::class)
         );
 
-        $manager = EntityManager::create(
-            $connection,
-            $configuration
-        );
+        $manager = $this->getEntityManager($settings, $connection, $configuration);
 
         $manager = $this->decorateManager($settings, $manager);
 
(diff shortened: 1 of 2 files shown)


guiwoda:
Are you sure extending the EntityManager is a good way of extending Doctrine?

matt-allan:
Unfortunately it is the only way to accomplish changes like [this](https://github.com/laravel-doctrine/orm/commit/bfa8e471238bddbbd603bf9c7e0d84dffd56216d).

Allowing the user to specify would allow users who prefer to use the default entity manager to continue doing so once master is tagged.

guiwoda:
Actually Doctrine [literally discourages it](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/EntityManager.php#L51)

patrickbrouwers:
I don't think it's a good idea to offer users the tools to extend Doctrine in this way. As @guiwoda pointed out Doctrine discourages it and offers a EntityManagerDecorator (which the config already supports).

The only reason why the EM is extended is that I couldn't find another way to painlessly implement the Collection hydration. But that's why it lives in the master branch, because I'm still looking into a better solution which has the same automatic hydration. A config setting to disable it was in the planning as well. But not based on defining the EntityManager class name, just a simple boolean value.

PR #73, opened 2015-10-20
Fix incorrect docblock (replaced JsonEncoder w/ ArrayEncoder)

maxbrokman:


maxbrokman:
StyleCI is picking up some stuff that aint me?

PR #78, opened 2015-10-28
resetManager() does not clear getManager()'s cache

sisve:
resetManager() does not clear getManager()'s cache (the managersMap array). This PR calls clears the cache so that a new manager is generated on the next call to getManager().


patrickbrouwers:
Thanks @sisve

FoxxMD:
:+1:

Issue #108, opened 2015-12-11
ENUM type error in doctrine:schema:validate


christiancannata:
Hi guys,
I get this error when I try to validate my schema:

"Unknown database type enum requested, Doctrine\DBAL\Platforms\MySQL57Platform may not support it."

I have added this in my doctrine.php config file:
 'custom_types'              => [
        'json' => LaravelDoctrine\ORM\Types\Json::class,
        'enum' => Doctrine\DBAL\Types\StringType::class
    ],

but this not work.

How can I fix it?

josenicomaia:
In order to solve it, I am using this snipet on AppServiceProvider.php:

``` php
    public function boot(\LaravelDoctrine\ORM\DoctrineManager $doctrineManager) {

        parent::boot();

        // Adding mapping server type to Doctrine type
        $doctrineManager->extendAll(function (\Doctrine\ORM\Configuration $configuration,
                \Doctrine\DBAL\Connection $connection,
                \Doctrine\Common\EventManager $eventManager) {

            $connection->getDatabasePlatform()->registerDoctrineTypeMapping('enum', 'string');
        });
    }
```

josenicomaia:
I guess that it will solve your problem @christiancannata

PR #111, opened 2015-12-20
Added support for configuring Mapping Types on a Connection.

josenicomaia:
Added support for configuring Mapping Types on a Connection.


josenicomaia:
#108 @christiancannata Now it will support using configuration ;D

PR #115, opened 2016-01-18
Move the binding of `validation.presence` into a deferred provider.

maxbrokman:
The Laravel implementation is now bound in a deferred provider, so cannot be overridden by a normal SP.

### DIFF
--- a/src/DoctrineServiceProvider.php
+++ b/src/DoctrineServiceProvider.php
@@ -31,7 +31,6 @@ use LaravelDoctrine\ORM\Exceptions\ExtensionNotFound;
 use LaravelDoctrine\ORM\Extensions\ExtensionManager;
 use LaravelDoctrine\ORM\Http\Middleware\BootExtensions;
 use LaravelDoctrine\ORM\Testing\Factory as EntityFactory;
-use LaravelDoctrine\ORM\Validation\DoctrinePresenceVerifier;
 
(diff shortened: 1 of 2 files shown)


patrickbrouwers:
Code style... AGAIN! ;) <3

maxbrokman:
GODAMN IT

maxbrokman:
I _think_ this will work, or at least doing it on my install worked. Had to deleted my services.php etc to get it functioning.

patrickbrouwers:
Okay, will test it when I'm home.

maxbrokman:
There is another option.

You could use [`Illuminate\Foundation\Application::getDeferredServices()` / `::setDeferredServices()`](https://github.com/laravel/framework/blob/0b185c99ba227570fefaf890c0ed764ac45a924c/src/Illuminate/Foundation/Application.php#L925) to grab the deferred list and pluck `validation.presence` out of it.

Seems a bit hacky though (to be fair this fix relies on the ordering of you service providers which ain't fun either) because the Container interface doesn't have knowledge of deferred / eager.

The problem is that `::make()` favours deferred providers over eager ones, by always loading the deferred provider immediately before passing through to the container. 

Otherwise I guess you'd have to get users to remove the validation provider which aint fun!

maxbrokman:
It works?!

:wink:

patrickbrouwers:
Nope :(

Issue #116, opened 2016-01-20
Getting error when trying to use Password Reset functionality in L5.2


glennjacobs:
I'm getting the following error in Laravel 5.2 (using version 1.1.0), which looks to be a bug...

```
ErrorException in AuthManager.php line 246:
call_user_func_array() expects parameter 1 to be a valid callback, class 'Illuminate\Auth\SessionGuard' does not have a method 'driver'

in AuthManager.php line 246
at HandleExceptions->handleError('2', 'call_user_func_array() expects parameter 1 to be a valid callback, class 'Illuminate\Auth\SessionGuard' does not have a method 'driver'', '/home/vagrant/homingpin-v2/vendor/laravel/framework/src/Illuminate/Auth/AuthManager.php', '246', array('method' => 'driver', 'parameters' => array()))
at call_user_func_array(array(object(SessionGuard), 'driver'), array()) in AuthManager.php line 246
at AuthManager->__call('driver', array()) in PasswordResetServiceProvider.php line 46
at AuthManager->driver() in PasswordResetServiceProvider.php line 46
at PasswordResetServiceProvider->LaravelDoctrine\ORM\Auth\Passwords\{closure}(object(Application), array()) in Container.php line 735
at Container->build(object(Closure), array()) in Container.php line 633
at Container->make('auth.password', array()) in Application.php line 674
at Application->make('auth.password') in Container.php line 1176
at Container->offsetGet('auth.password') in Facade.php line 151
```

maxbrokman:
@patrickbrouwers looks like this 

https://github.com/laravel-doctrine/orm/blob/f11272540167d5e8b72cdcf61ddb5f56cad823c8/src/Auth/Passwords/PasswordResetServiceProvider.php#L46

``` php
    /**
     * Register the password broker instance.
     * @return void
     */
    protected function registerPasswordBroker()
    {
        $this->app->singleton('auth.password', function ($app) {
            // The password token repository is responsible for storing the email addresses
            // and password reset tokens. It will be used to verify the tokens are valid
            // for the given e-mail addresses. We will resolve an implementation here.
            $tokens = $app['auth.password.tokens'];

            $users = $app['auth']->driver()->getProvider();

            $view = $app['config']['auth.password.email'];

            // The password broker uses a token repository to validate tokens and send user
            // password e-mails, as well as validating that password reset process as an
            // aggregate service of sorts providing a convenient interface for resets.
            return new PasswordBroker(
[...]

PR #117, opened 2016-01-21
Replace direct binding of `auth.password.broker` with our own implementation of its manager class.

maxbrokman:
This is a potential fix for #116 

Sorry @patrickbrouwers but I've not done any end-end testing on this, I've still not got a test installation setup for this sort of stuff and we don't use passwords.

 In 1.1 the definition of the password reset broker is not compatible with Laravel 5.2. As part of fixing this I've decided to extend Laravel's `PasswordBrokerManager` so that we can bind our own token repository into it and then use it as is.

 Extending the manager seems fairly fragile, but there's no other way of forcing a different token repository in. I'm considering doing something similar as an upstream pull request to Laravel which means we wouldn't have to rely on extending this class.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,9 +3,11 @@ composer.phar
 composer.lock
 .DS_Store
 .php_cs.cache
+/coverage
 
 /tests/Stubs/storage/framework/views/*
 !/tests/Stubs/storage/framework/views/.gitkeep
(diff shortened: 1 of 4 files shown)


patrickbrouwers:
Replaced by 73ed9ffb812a59f2889d1b17cca35d2b1e1495c0

PR #125, opened 2016-02-12
Exporter is now able to deal with entity namespaces. 

guiwoda:
Also added a test for a simple use case (only an id).


patrickbrouwers:
:+1:

PR #128, opened 2016-03-07
[FIX] Fixing password broker which was returning NULL instead of the entity…

glennjacobs:
The password broker wasn't getting a valid entity manager when creating the DoctrineTokenRepository object.

I've updated the class to use Doctrine\ORM\EntityManagerInterface which appears to have solved the problem.

### DIFF
--- a/src/Auth/Passwords/PasswordBrokerManager.php
+++ b/src/Auth/Passwords/PasswordBrokerManager.php
@@ -2,7 +2,7 @@
 
 namespace LaravelDoctrine\ORM\Auth\Passwords;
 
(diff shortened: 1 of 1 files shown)


glennjacobs:
Actually this is wrong. The problem is I made my own PasswordReminder entity, which isn't going to work.

patrickbrouwers:
Shouldn't be necessary if you include the provided ServiceProvider right?

glennjacobs:
Turned out to be yet another issue with service providers being in the wrong order... I was barking up the wrong tree.

PR #129, opened 2016-03-09
[FIX] Add exit codes to commands

driesvints:
- Add exit codes for commands so third party tools can handle them

At the moment these commands are failing silently because external tools don't know when a command failed or succeeded. By returning an exit code (> 0) tools like CI tools etc can handle these commands properly.

To give a very specific example, we use CircleCI to check the schema validate command to see if it produced any SQL so we know if our mappings are ok. By adding the exit codes CircleCI knows when the command failed or succeeded.

I've tried porting the logic from the Doctrine commands as best as I could. Because you're using multiple managers, the implementation may vary slightly. 

I'm not really sure if this is "breaking" or not. People using these commands in tools like CircleCI may experience different behaviour in those tools when this gets merged. Builds may suddenly fail because a different exit code is returned. However, I do think this is preferred behaviour.

Thanks for considering!

### DIFF
--- a/src/Console/EnsureProductionSettingsCommand.php
+++ b/src/Console/EnsureProductionSettingsCommand.php
@@ -43,7 +43,7 @@ class EnsureProductionSettingsCommand extends Command
                 $this->error('Error for ' . $name . ' entity manager');
                 $this->error($e->getMessage());
 
-                return;
+                return 1;
             }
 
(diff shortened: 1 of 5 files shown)


maxbrokman:
Thanks for this, something I noticed was missing when we added `schema:validate` to our Travis build.

patrickbrouwers:
Looks good to me and shouldn't break any behaviour.

driesvints:
Whohoo! 🎉

PR #134, opened 2016-03-22
[FEATURE] Add support for PHP metadata driver

zorji:
Add support for PHP metadata driver

http://doctrine-orm.readthedocs.org/en/latest/reference/php-mapping.html#php-files

### DIFF
--- a/config/doctrine.php
+++ b/config/doctrine.php
@@ -12,7 +12,7 @@ return [
     | paths setting to the appropriate path and replace App namespace
     | by your own namespace.
     |
-    | Available meta drivers: fluent|annotations|yaml|xml|config|static_php
+    | Available meta drivers: fluent|annotations|yaml|xml|config|static_php|php
     |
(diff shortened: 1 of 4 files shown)


src/Extensions/MappingDriverChain.php
```diff
@@ -39,7 +40,7 @@ public function addPaths(array $paths = [])
 
         if ($driver instanceof AnnotationDriver || $driver instanceof StaticPHPDriver) {
             $driver->addPaths($paths);
-        } elseif ($driver instanceof FileDriver) {
+        } elseif ($driver instanceof FileDriver || $driver instanceof PHPDriver) {
```
patrickbrouwers: Not needed, PHPDriver extends FileDriver

patrickbrouwers:
If you could fix my comment, it's good to go

patrickbrouwers:
:+1:

PR #135, opened 2016-03-28
[FIX] doctrine generate entities command

amitailanciano:
- change default path to current working directory and use argument path if provided
- use same defaults as original doctrine command

### DIFF
--- a/src/Console/GenerateEntitiesCommand.php
+++ b/src/Console/GenerateEntitiesCommand.php
@@ -52,7 +52,7 @@ class GenerateEntitiesCommand extends Command
             $metadatas = $cmf->getAllMetadata();
             $metadatas = MetadataFilter::filter($metadatas, $this->option('filter'));
 
-            $destPath = base_path('app/Entities');
+            $destPath = base_path($this->argument('dest-path') ? $this->argument('dest-path') : '.');
 
             if (!is_dir($destPath)) {
                 mkdir($destPath, 0777, true);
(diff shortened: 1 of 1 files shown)


amitailanciano:
maybe default entity path could be set in config and use that instead of current working directory if not supplied

patrickbrouwers:
Good spot, forgot to put that back to dest-path.

But why `'.'` when no dest-path is given?

patrickbrouwers:
Going to close this one in favour of https://github.com/laravel-doctrine/orm/pull/133

amitailanciano:
that's fine, but can you also include the parameter defaults from the original command?

This commit: https://github.com/laravel-doctrine/orm/pull/135/commits/96ed3e16f4269b23f9ddded98736499ab2b874a3

If you'd like to implement that differently that's ok, I just think it's really helpful for those to be the defaults

patrickbrouwers:
It already has those defaults:

if(null) == if(false)

amitailanciano:
I was being explicit with those two -- you are right that they are not needed, but there are also two conditions where `true` is the default if `null`

amitailanciano:
See https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Tools/Console/Command/GenerateEntitiesCommand.php for the original command defaults

patrickbrouwers:
Ah didn't spot the two true's. If you can make a new PR with that commit, would be nice, thanks.

amitailanciano:
Here you go https://github.com/laravel-doctrine/orm/pull/138

PR #138, opened 2016-04-06
[FIX] add default argument values from original doctrine generate entities …

amitailanciano:
- add default argument values from original doctrine generate entities command


patrickbrouwers:
:+1:

PR #145, opened 2016-04-29
[FIX] Mapping convertion using database

R3VoLuT1OneR:
Hey I just found small "syntax" errors in "doctrine:convert:mapping" command.

Please merge the fix.

### DIFF
--- a/src/Console/ConvertMappingCommand.php
+++ b/src/Console/ConvertMappingCommand.php
@@ -3,6 +3,7 @@
 namespace LaravelDoctrine\ORM\Console;
 
 use Doctrine\Common\Persistence\ManagerRegistry;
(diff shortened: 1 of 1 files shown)


patrickbrouwers:
Please re-add the `=` to `--from-database`

R3VoLuT1OneR:
@patrickbrouwers If remove it, code `$this->option('from-database') === true` never be true.

`--from-database` - Without `=` parsed as boolean ( set or not set ) and `--form-database=` parsed always as string. Because here we have option set/not set and no value required I think should be used `--form-data`.

R3VoLuT1OneR:
@patrickbrouwers And in original doctrine command it is `--from-database` without `=`

patrickbrouwers:
Ah I was in the impression you could pass the database name to that option.

Issue #151, opened 2016-06-07
[BUG] Configuration for mappings path ignored by mappingsImportCommand


Jacco-V:
### Version Info

Package v1.1.10,
Laravel v5.2.
### Expected behaviour

Having specified my `config/doctrine.php` file as follows (petty much the unchanged from the default):

```
return [
    'managers'                  => [
        'default' => [
            'dev'        => env('APP_DEBUG'),
            'meta'       => env('DOCTRINE_METADATA', 'SimplifiedXml'),
            'connection' => env('DB_CONNECTION_DEFAULT', 'default'),
            'namespaces' => [
                'App\Entities'
            ],
            'paths'      => [
                base_path('database/mappings') => 'App\Entities',
            ],

            // (...)
        ]
    ]
];
```

I would expect the console command: `php artisan doctrine:mapping:import` to write the mapping information into the directory: `database/mappings/NameOfTable.orm.xml`
### Actual behaviour

Because in [MappingImportCommand.php:43](https://github.com/laravel-doctrine/orm/blob/1.1.10/src/Console/MappingImportCommand.php#L43) the default is hardcoded as `'app/Mappings'`, the mapping information is written to `app/Mappings/NameOfTable.orm.xml` instead.
### Steps to reproduce the behaviour

Create a new project, include laravel-doctrine with its default configuration, add some table to your standard database, run console command `php artisan doctrine:mapping:import`

patrickbrouwers:
Please make a PR for these changes.

patrickbrouwers:
Closing this for now (see PR). Might revisit this in the future (major release). For now I think using the {dest-path} argument is good enough.

Issue #143, opened 2016-04-14
[BUG] Cannot use entity files or create entity files from imported mappings files


jaccoTopVintage:
### Expected behaviour

```
$> php artisan doctrine:generate:entities /

Generating getter and setter for default entity manager...
Processing entity "App\Entities\Nnnnnnnn"

Entity classes generated to "D:\Web\TopManager"
```
### Actual behaviour

```
$> php artisan doctrine:generate:entities /

Generating getter and setter for default entity manager...

  [Doctrine\Common\Persistence\Mapping\MappingException]
  Invalid mapping file 'App.Entities.Nnnnnnnn.orm.xml' for class 'App\Entities\Nnnnnnnn'.
```
### Steps to reproduce the behaviour
- Setup a new Laravel project.
- In the configuration file `config/doctrine.php` make sure the following lines are set as shown below:

```
'managers' => [
    'default' => [
        'meta' => 'SimplifiedXml',
        'paths' => [
            base_path('database/mappings') => 'App\Entities',
        ],
    ],
];
```
- Create the following additional file structure:

```
[project]
   +- [App]
       +- [Entites]
   +- [database]
       +- [mappings]
```
- create mappings files from an existing database using console command:  
  `php artisan doctrine:mapping:import xml database/mappings`
- attempt to generate entities from the imported mappings files using console command:  
  `php artisan doctrine:generate:entities /`

jaccoTopVintage:
## Additional information

I think there needs to be a change to a Doctrine class to accommodate the use of mapping files (`.xml`/`.yml`) for Doctrine when used in combination with Laravel.

More specifically, in file [`Doctrine/Common/Persistence/Mapping/Driver/FileDriver.php`](https://github.com/doctrine/common/blob/v2.6.1/lib/Doctrine/Common/Persistence/Mapping/Driver/FileDriver.php):

**Current lines of code**

```
116: if (!isset($result[$className])) {
117:     throw MappingException::invalidMappingFile($className, str_replace('\\', '.', $className) . $this->locator->getFileExtension());
118: }
119:
120: return $result[$className];
```

**Required lines of code**

```
116: $classNameBase = explode('\\', $className);
117: $classNameBase = end($classNameBase);
118: if (!isset($result[$classNameBase])) {
119:     throw MappingException::invalidMappingFile($className, $classNameBase . $this->locator->getFileExtension());
120: }
121: return $result[$classNameBase];
```

<br>
[...]

PR #155, opened 2016-06-09
[FIX] Use configuration defined path instead of hard-coded value for defaul…

Jacco-V:
Use configuration defined path instead of hard-coded value for default destination path for writing imported mapping-files. see [issue #151](https://github.com/laravel-doctrine/orm/issues/151)

### DIFF
--- a/src/Console/GenerateEntitiesCommand.php
+++ b/src/Console/GenerateEntitiesCommand.php
@@ -51,8 +51,8 @@ class GenerateEntitiesCommand extends Command
             $cmf->setEntityManager($em);
             $metadatas = $cmf->getAllMetadata();
             $metadatas = MetadataFilter::filter($metadatas, $this->option('filter'));
-
-            $destPath = base_path($this->argument('dest-path') ?:'app/Entities');
+			
+            $destPath = base_path($this->argument('dest-path') ? $this->argument('dest-path') : current(config("doctrine.managers.{$name}.paths")));
 
             if (!is_dir($destPath)) {
                 mkdir($destPath, 0777, true);
--- a/src/Console/MappingImportCommand.php
+++ b/src/Console/MappingImportCommand.php
@@ -40,7 +40,7 @@ class MappingImportCommand extends Command
         $emName = $this->option('em');
         $em     = $registry->getManager($emName);
 
-        $destPath = base_path($this->argument('dest-path') ? $this->argument('dest-path') : 'app/Mappings');
+        $destPath = $this->argument('dest-path') ? base_path($this->argument('dest-path')) : key(config("doctrine.managers.{$emName}.paths"));
 
         $type = $this->argument('mapping-type');
 


patrickbrouwers:
Why `current()` for one and `key()` for the other?

Jacco-V:
Well, because of how the configuration directive is designed.

if you use the supplied [default path configuration](https://github.com/laravel-doctrine/orm/blob/1.1.10/config/doctrine.php#L31):

``` php
  'paths'      => [
      base_path('app')
  ],
```

and run command:
`php artisan doctrine:generate:entities`

you get an error:

```
  [Doctrine\Common\Persistence\Mapping\MappingException]
  File mapping drivers must have a valid directory path, however the given path 0 seems to be incorrect!
```

In other words, the command above expects to read from the `key` of the `'paths'` array.

so, to fix the issue, you change the defaults to:

``` php
  'paths'      => [
      base_path('app') => 'yyyyy'
  ],
```

However, when you then try command `php artisan doctrine:generate:proxies`, you'll get the error:

```
  [Doctrine\Common\Persistence\Mapping\MappingException]
  Class 'yyyyy\Mappings\Users' does not exist
```

So, in order to supply both the commands above with the configuration values they expect, you need a path configuration directive that looks like:

``` php
'paths' => [
    base_path('database/mappings') => 'App\Entities',
],
```

which is, in result, why you need the `doctrine:mapping:import` command to read the `key` of the mappings array, and the `doctrine:generate:entities` command to read the value of the path configuration directive, hence the `current()`.

Of course, I now realise that it would be a good idea to change the directive in [the default doctrine configuration file](https://github.com/laravel-doctrine/orm/blob/1.1.10/config/doctrine.php) as well.

If you want, I'll add the changes to the default Doctrine config file to this pull request.

patrickbrouwers:
Hm, I'm not really happy with that. It will break all existing apps, as the `paths` array should be an array of values and not keys. Why can't both commands not just take the first value of the `paths` array? It would keep everything working the way it is.

Jacco-V:
It would not break any existing apps, because the default configuration file is only exported once, upon installation, by issuing the command: `php artisan vendor:publish --tag="config"`.

> Why can't both commands not just take the first value of the paths array?

Well, there are two paths: 
- the path to the mappings
- the path to the entities

Since they don't live in the same directory (nor would I want them to live in the same directory), you need two paths, and a method to tell which one is which. Given those requirements, the solution now used by the package is (although arguably not the cleanest solution) workable.

If you would like to change the use of a `key=>value` pair into something else, you would have to coordinate a cross-package change, since the format in which the path configuration directive has to be specified, comes from the [Doctrine](https://github.com/doctrine/doctrine2) package. Laravel-Doctrine is only supplying the expected format.

guiwoda:
@Jacco-V The `paths` configuration is only meant for **mappings** paths. The only case when mappings path match entities path is on **Annotations**, because that's the only driver that cares about where your Entities are. The rest of the drivers don't need to know that, so the only path they'll need is where mappings are.

The entity generator needs an entity path. As we can't assume any specific driver, we shouldn't be reusing the `paths` configuration for it (that would only make sense on Annotations). So, if we want to get rid of the hard-coded `app/Entities` value, we can either:
- require a path parameter: makes calls more verbose if `app/Entities` was enough, breaks previous signature (I don't care a lot about this commands signature anyway)
- add an additional config key: will have to check existance because previous imports won't have it. Doesn't make sense in annotations.

I feel that neither is worth the effort.

I'd only modify the mapping generator to use `paths` as it is right now, but I wouldn't modify the entity generator at all.

Jacco-V:
**1.** If it is only meant for the mappings path, then there is a bug somewhere in `doctrine:generate:proxies`, because it clearly uses the same configuration directive.

**2.** I don't see how only the annotations would, on the one hand be the only driver that needs to know where my entities are, while on the other hand a path directive to the entities would not make sense when using annotations (as per your comment). What am I missing?

guiwoda:
@Jacco-V 
We might be looking at this the wrong way. I think the problems revolve around the `SimplifiedXml` mapping driver and how other drivers expect an array of `paths`, while `SimplifiedXml` expects an associative array of `path => NS-prefix`.

Jacco-V:
The various drivers and their simplified counterparts (`XML` vs `SimplifiedXml`, etc) certainly don't make it easier to get the configuration working.

Yet, it seems like an inconsistency for the `SimplifiedXml` driver to read the `namespace prefix` from the `path` directive, while there is also a [`namespace` directive](https://github.com/laravel-doctrine/orm/blob/1.1.10/config/doctrine.php#L29) in the same configuration file?

patrickbrouwers:
That's what I meant with breaking changes. I've never used `SimplifiedXml` so didn't know it was the other way around, which now does explain the way you solved it. 
But I certainly dont' think it's a good way to let the `SimplifiedXml` drive the way the import commands work :)

Jacco-V:
I presumed that the different drivers would not require conflicting configuration settings, but apparently, this assumption was incorrect.

Since I did not use the other drivers (annotations, yaml, etc) I don't know which incompatibilities the proposed changes would create. At the same time, I _do_ feel that a fix is needed, because right now, it is _really_ hard to get the various commands to works (using the `SimplifiedXML` driver, I don't know about others) and I'm not sure I would have figured it out in any reasonable time frame, without the example configuration in issue [#143 ](https://github.com/laravel-doctrine/orm/issues/143)

patrickbrouwers:
I don't really agree with the fact that it's hard to get it working, as you can perfectly fine pass the `dest-path`in those commands and set it to the directory you want. It might not be the most optimal way, but it does work.

Jacco-V:
It is hard to get working, because the default configuration causes errors.

patrickbrouwers:
I think in cases like these you should always refer to the docs of the mapping driver: http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/reference/xml-mapping.html
As we just pass the `paths` array value to the driver the contents of it should match the expected value of the mapping driver. I don't think there's really anything this package can do about it, as it's an (intentional) inconsistency in the parent package.

Jacco-V:
The configuration layout/arrangement of the Laravel-Doctrine package differs from the parent package.
The parent package has expectations on how the configuration looks. Those expectations apparently differ from driver to driver.
The laravel-doctrine package provides the parent package with the configuration directives it needs. 

Then why would it not be possible for the laravel-doctrine package to transform the local configuration syntax into a configuration syntax as required by the parent package, based on the driver specified?!

patrickbrouwers:
It is possible with a little effort of the developer using the package... If you are using SimplifiedXml driver you must have read their docs and now that they expect another format than given in the doctrine.php config. I don't see any responsibility in this package for that, except maybe a reference in our docs to clarify that.

Jacco-V:
We clearly have different opinions on user friendliness and good package design.

patrickbrouwers:
Dude, it's such a trivial change in the config file, why would we make it more complex than it needs to be...? If you don't agree with the inconsistency, this is not the place to complain about it: it was not our decision. We support SimplifiedXml perfectly fine, by letting the user change the config file. User friendly enough. I really don't get why we get the blame for bad package design, where it's clearly due to the fact the parent package decided to do it different.

patrickbrouwers:
To get the facts straight:
- The current PR will not work for the majority of the package users, as it uses the config structure of SimplifiedXml, and therefor not a solid solution.
- The current `paths` config value doesn't need to and shouldn't be changed as it would be a breaking change for the majority of the users. Support for SimplifiedXml is present, by letting the user enter the correct format necessary for that specific driver. This package doesn't need to be responsible for translating this format.
- Auto-selecting the correct path from the config for these commands is a nice-to-have, not a necessity. It's perfectly workable (just like it works for standalone usage outside) by passing the `--dest-path`  option/argument.

maxbrokman:
I don't think attempting to provide a default path for entities / mappings based on the `paths` config is sensible, it's a bit magic and it's not what the `paths` config if for.

It also seems like the PR as is would break the default behaviour of the mapping import command for anyone who is not using SimplifiedXML (so 99% of people using this package probably) and has not changed their config file. Worse the error generated will be extremely opaque to most people (`0 is not a path`).

The default suggested path are very Laravel-esque (see all the `artisan make` commands), which is fine as it's a Laravel package. There is an argument to be made for removing the default and forcing the user to provide `dest-path`, a breaking change but a small one given this is a manual command presumably run infrequently. Making such a change depends on how strictly @patrickbrouwers wants to follow SemVer.

For users who want to use SimplifiedXML the correct solution would be to document the change required to the `paths` config, both on the site and probably briefly inline in the config file itself.

I'm against the PR.

tldr: PR as is will break most installs to no clear benefit.

patrickbrouwers:
Config difference cleared up by: https://github.com/laravel-doctrine/docs/commit/e5658c1520b7bef0ceaa2590875ff8fd449505d7

Jacco-V:
@patrickbrouwers This [comment of mine](https://github.com/laravel-doctrine/orm/pull/155#issuecomment-226795485) earlier was pretty snarly. There was no need for that. I apologize.

PR #158, opened 2016-06-16
[FEATURE] Split service provider

adridev:
Move PresenceVerifierProvider to its own Service provider. This Service provider allows to register it independently to avoid override the native validation.presence service by default.


xavierhb:
+1 👍

patrickbrouwers:
`PresenceVerifierProvider::class` is already a ServiceProvider, so it wouldn't require creating a new ServiceProvider. 

However removing the auto-register from the main service provider in version 1.1 is not possible as it will break all existing installs. 
We could only do this in a v2 release which is not really in the planning. The only non-breaking way to handle this is by adding a config setting which can turn on/off the doctrine validation.
In the current state I can't merge this PR. Feel free to join us on Slack if you want to discuss this more.

adridev:
Ok, good points. I'll review the toggle option in config.

Thanks, @patrickbrouwers .

patrickbrouwers:
Thanks :)

PR #160, opened 2016-06-16
[FEATURE] toggle PresenceVerifierProvider based on configuration

adridev:
Toggle service registration based on the configuration.

### DIFF
--- a/config/doctrine.php
+++ b/config/doctrine.php
@@ -171,5 +171,6 @@ return [
     */
     'gedmo'                     => [
         'all_mappings' => false
-    ]
+    ],
+    'doctrine_presence_verifier' => true,
 ];
--- a/config/doctrine.php
+++ b/config/doctrine.php
@@ -172,5 +172,13 @@ return [
     'gedmo'                     => [
         'all_mappings' => false
     ],
+    /*
+     |--------------------------------------------------------------------------
+     | Validation
+     |--------------------------------------------------------------------------
+     |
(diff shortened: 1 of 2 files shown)


src/DoctrineServiceProvider.php
```diff
@@ -309,10 +310,14 @@ protected function isLumen()
...
+     * Checks if configuration key "doctrine_presence_verifier" is defined and returns its value (true|false), if it
+     * doesn't returns true by default
+     *
      * @return bool
      */
-    protected function registerPresenceValidator()
+    protected function shouldRegisterDoctrinePresenceValidator()
     {
-        return (null !== @$this->app['config']['doctrine']['doctrine_presence_verifier']) ? $this->app['config']['doctrine']['doctrine_presence_verifier'] : true;
+        $configValue = @$this->app['config']['doctrine']['doctrine_presence_verifier'];
```
patrickbrouwers: `$this->app['config']->get('doctrine.doctrine_presence_verifier', true)` is more reliable

config/doctrine.php
```diff
@@ -171,5 +171,6 @@
     */
     'gedmo'                     => [
         'all_mappings' => false
-    ]
+    ],
+    'doctrine_presence_verifier' => true,
```
patrickbrouwers, line 183: Could you add a comment block on top of this with the explanation of what it does?

maxbrokman:
good work buddy

PR #161, opened 2016-06-19
Fix the bug "Invalid PathExpression" in validation service

globtec:
[FIX] "Invalid PathExpression" in validation service

Showed when the validation service was used with Doctrine version 2.5.4.
In the expression COUNT, was used the entity alias.
### Usage:

``` php
$this->validate($requrest, [
    'username' => 'unique:Model\Entity\User,username'
]);
```


patrickbrouwers:
@globtect Could you fix the failing test, thanks.

globtec:
Sorry, ignore this pull request, my problem is when I use a oneToOne relation as primary key, the following dql request fail: _[Semantical Error] line 0, col 13 near 'a) FROM MyEntity': Error: Invalid PathExpression. Must be a StateFieldPathExpression_

PR #168, opened 2016-07-14
[FIX] Register proxy autoloader

R3VoLuT1OneR:
I just found that `laravel-doctrine` not registering proxy autoloader.
Problem that I got is that on unserializing doctrine entity in laravel queue for example, proxy classes not loaded/included so I got next exception on trying to `merge` unserialized entity:

```
Doctrine\ORM\Mapping\MappingException: Class "__PHP_Incomplete_Class" is not a valid entity or mapped super class. in vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/MappingException.php:346
```

P.S. Implementation not perfect but works for me.
P.S.S. Don't have time and idea how to test it.

### DIFF
--- a/src/DoctrineServiceProvider.php
+++ b/src/DoctrineServiceProvider.php
@@ -3,6 +3,7 @@
 namespace LaravelDoctrine\ORM;
 
 use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\Common\Proxy\Autoloader;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadataFactory;
@@ -70,6 +71,7 @@ class DoctrineServiceProvider extends ServiceProvider
         $this->registerPresenceVerifier();
         $this->registerConsoleCommands();
         $this->registerCustomTypes();
+        $this->registerProxyAutoloader();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


guiwoda:
@R3VoLuT1OneR can you describe your scenario? I've been using LD with and without autogenerating proxies and never had a problem with proxy autoloading

R3VoLuT1OneR:
@guiwoda We using event listener with `ShouldQueue` interface and event class with doctrine entity property.

For example:

```
class CreateAccountEvent {
   public function __construct(App\Entity\Account $account) {
       $this->account = $account;
   }
}
```

This event object serialized and de-serialized in the queue worker.
After de-serialization we making `\EntityManager::merge($account)` and it is falling with next exeption:

```
Doctrine\ORM\Mapping\MappingException: Class "__PHP_Incomplete_Class" is not a valid entity or mapped super class.
```

After small debug I realized that that namespace `\DoctrineProxies` not autoloaded that why `unserialize` falling leaving failed objects as `__PHP_Incomplete_Class` class.

patrickbrouwers:
I personally think you should never serialize your entities on the queue. It's a lot better to refetch the entity in the listener with all the information that's needed.

maxbrokman:
Don't see the harm in supporting that use case though, even if it's not ideal?

R3VoLuT1OneR:
@patrickbrouwers I think you right, but still `laravel-doctrine\orm` should support unserializing entities as Doctrine2 supports it.

I got problems when tried to load entity with `\EntityManager::find(...)` and then `\EntityManager::merge(...)` and `\EntityManager::refresh(..)` helped.
We using default Laravel DB engine and Doctrine2 (migrating to Doctrine) so maybe was some caching or another problem. Gonna investigate the my real problem tomorrow.

src/DoctrineServiceProvider.php
```diff
@@ -222,6 +224,24 @@ protected function extendAuthManager()
     }
 
     /**
+     * Register proxy autoloader
+     *
+     * @return void
+     */
+    public function registerProxyAutoloader()
+    {
+        $this->app->afterResolving(IlluminateRegistry::class, function (ManagerRegistry $registry) {
```
patrickbrouwers: can you change IlluminateRegistry -> ManagerRegistry. Thanks.

R3VoLuT1OneR:
@patrickbrouwers Changed `IlluminateRegistry` to `ManagerRegistry`

patrickbrouwers:
Can you retarget this to `1.2` please. `develop` is a year behind.

R3VoLuT1OneR:
Created new pull request on `1.2` branch #172

patrickbrouwers:
Thanks

PR #181, opened 2016-09-25
[FIX] Fix typehint mistake on getEntity method in DoctrineChannel + add toArray method to DoctrineChannel to match Laravel behaviour.

leeovery:
- Typehint on getEntity method in DoctrineChannel was wrong. Was set to the LD Notification MappedSuperclass, but should have been the \Illuminate\Notifications\Notification class. You had aliased that to LaravelNotification.
- Added toArray check to match Laravel behaviour for database channel.

### DIFF
--- a/src/Notifications/DoctrineChannel.php
+++ b/src/Notifications/DoctrineChannel.php
@@ -57,10 +57,12 @@ class DoctrineChannel
             return $notification->toEntity($notifiable);
         } elseif (method_exists($notification, 'toDatabase')) {
             return $notification->toDatabase($notifiable);
+        } elseif (method_exists($notification, 'toArray')) {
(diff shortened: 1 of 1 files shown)


patrickbrouwers (changes requested):
I intentionally left of the toArray method as it would indicate another behaviour than is expected by the Channel.

leeovery:
Didnt think of that. Removed that part and left the typehint fix in. I meant to do that in two separate PR's.

patrickbrouwers (approved):
👍

patrickbrouwers (changes requested):
Could you fix the CS, thanks. (See StyleCI)

patrickbrouwers (approved):

PR #186, opened 2016-10-12
[FEATURE] Different EntityManager class

tomkoreny:
# Configurable EM class

I came across this issue when I was trying to use extended EM class by [Kdyby](https://github.com/Kdyby/Doctrine). There is nothing much to say about it, I believe this wouldn't hurt anyone, but it opens up the possibility to extend standart doctrine classes a bit. 
- Instead of hard coded entity manager class use config value
- Added this value to config

### DIFF
--- a/config/doctrine.php
+++ b/config/doctrine.php
@@ -32,6 +32,7 @@ return [
                 base_path('app')
             ],
             'repository'    => Doctrine\ORM\EntityRepository::class,
+            'entity_manager'=> Doctrine\ORM\EntityManager::class,
             'proxies'       => [
                 'namespace'     => false,
                 'path'          => storage_path('proxies'),
--- a/src/EntityManagerFactory.php
+++ b/src/EntityManagerFactory.php
@@ -120,7 +120,7 @@ class EntityManagerFactory
 
         $configuration->setEntityListenerResolver($this->resolver);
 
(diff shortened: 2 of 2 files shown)


maxbrokman:
I'm not sure why you want to use a different EM but the class you've linked goes against how the EM is [supposed to be used](https://github.com/doctrine/doctrine2/blob/a353cb81a34e70189ad3246d07c844d7fe8699ce/lib/Doctrine/ORM/EntityManager.php#L52-L55)

> You should never attempt to inherit from the EntityManager: Inheritance is not a valid extension point for the EntityManager. Instead you should take a look at the [{@see \Doctrine\ORM\Decorator\EntityManagerDecorator}](https://github.com/doctrine/doctrine2/blob/a353cb81a34e70189ad3246d07c844d7fe8699ce/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php) and wrap your entity manager in a decorator.

LD supports such decorators, simply add a `decorator` key to your config and it will get [picked up](https://github.com/laravel-doctrine/orm/blob/dcaae65e87b987c855f3f3d34c3194ae767f8765/src/EntityManagerFactory.php#L371-L388). It's likely you'll be able to achieve what you want through use of a decorator but I couldn't say without knowing more about your use case.

PR #191, opened 2016-10-24
[FIX] Fluent exporter improvements

FoxxMD:
- Put single quote in correct position when writing custom repository
- Fix typo on 'unique' node array accessor for joinColumns
- Discern between bi-directional 1-to-1 associations (fix accessing non-existent joinColumns node)
- Make exported file names match generated class names (psr-4 compatibility)

### DIFF
--- a/src/Console/Exporters/FluentExporter.php
+++ b/src/Console/Exporters/FluentExporter.php
@@ -13,7 +13,7 @@ class FluentExporter extends AbstractExporter
      */
     protected $_extension = '.php';
 
-	protected $_mappingClassName;
+    protected $_mappingClassName;
 
     /**
      * @const
@@ -25,7 +25,7 @@ class FluentExporter extends AbstractExporter
      */
     public function exportClassMetadata(ClassMetadataInfo $metadata)
     {
-        $lines   = [];
+        $lines = [];
         $lines[] = '<?php';
         $lines[] = null;
         $lines[] = 'namespace App\Mappings;';
@@ -36,7 +36,7 @@ class FluentExporter extends AbstractExporter
         }
 
         $lines[] = null;
-        $lines[] = 'class ' . $this->getClassNameShort($metadata) . 'Mapping extends ' . $this->getExtendShort($metadata) . ' {';
+        $lines[] = 'class '.$this->getClassNameShort($metadata).'Mapping extends '.$this->getExtendShort($metadata).' {';
         $lines[] = null;
 
         foreach ($this->exportMapFor($metadata) as $line) {
@@ -54,28 +54,28 @@ class FluentExporter extends AbstractExporter
         return implode("\n", $lines);
     }
 
-	public function export()
-	{
-		if ( ! is_dir($this->_outputDir)) {
-			mkdir($this->_outputDir, 0775, true);
-		}
+    public function export()
+    {
+        if (!is_dir($this->_outputDir)) {
+            mkdir($this->_outputDir, 0775, true);
+        }
 
-		foreach ($this->_metadata as $metadata) {
-			// In case output is returned, write it to a file, skip otherwise
-			if($output = $this->exportClassMetadata($metadata)){
-				$path = $this->_outputDir . '/' . str_replace('\\', '.', $this->_mappingClassName) . $this->_extension;
-				$dir = dirname($path);
-				if ( ! is_dir($dir)) {
-					mkdir($dir, 0775, true);
-				}
-				if (file_exists($path) && !$this->_overwriteExistingFiles) {
-					throw ExportException::attemptOverwriteExistingFile($path);
-				}
-				file_put_contents($path, $output);
-				chmod($path, 0664);
-			}
-		}
-	}
+        foreach ($this->_metadata as $metadata) {
+            // In case output is returned, write it to a file, skip otherwise
+            if ($output = $this->exportClassMetadata($metadata)) {
+                $path = $this->_outputDir.'/'.str_replace('\\', '.', $this->_mappingClassName).$this->_extension;
+                $dir = dirname($path);
+                if (!is_dir($dir)) {
+                    mkdir($dir, 0775, true);
+                }
+                if (file_exists($path) && !$this->_overwriteExistingFiles) {
+                    throw ExportException::attemptOverwriteExistingFile($path);
+                }
+                file_put_contents($path, $output);
+                chmod($path, 0664);
+            }
+        }
+    }
 
     /**
      * @param mixed $var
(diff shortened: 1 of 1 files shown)


src/Console/Exporters/FluentExporter.php
```diff
@@ -51,6 +55,29 @@ public function exportClassMetadata(ClassMetadataInfo $metadata)
...
+    public function export()
+    {
+        if (!is_dir($this->_outputDir)) {
+            mkdir($this->_outputDir, 0775, true);
+        }
+
+        foreach ($this->_metadata as $metadata) {
+            // In case output is returned, write it to a file, skip otherwise
+            if ($output = $this->exportClassMetadata($metadata)) {
+                $path = $this->_outputDir.'/'.str_replace('\\', '.', $this->_mappingClassName).$this->_extension;
```
maxbrokman, line 67: is this the convention for fluent mappers? like `Namespace.Class` for `Namespace\Class` mappings?

src/Console/Exporters/FluentExporter.php
```diff
@@ -414,44 +453,44 @@ private function exportAssociations(array $associationMapping)
...
             foreach ($joinColumns as $c) {
-                $columnName           = $c['name'];
+                $columnName = $c['name'];
                 $referencedColumnName = $c['referencedColumnName'];
-                $nullable             = !isset($c['nullable']) || $c['nullable'] == true ? 'true' : 'false';
-                $unique               = !isset($c['unique']) || $c['nullable'] == false ? 'false' : 'true';
-                $onDelete             = isset($c['onDelete']) ? '\'' . $c['onDelete'] . '\'' : 'null';
-                $columnDefinition     = isset($c['columnDefinition']) ? '\'' . $c['columnDefinition'] . '\'' : 'null';
+                $nullable = !isset($c['nullable']) || $c['nullable'] == true ? 'true' : 'false';
+                $unique = !isset($c['unique']) || $c['unique'] == false ? 'false' : 'true';
```
maxbrokman, line 471: good catch, copy paste attacks again

src/Console/Exporters/FluentExporter.php
```diff
@@ -414,44 +453,44 @@ private function exportAssociations(array $associationMapping)
...
         if ($joinColumns) {
             foreach ($joinColumns as $c) {
-                $columnName           = $c['name'];
+                $columnName = $c['name'];
                 $referencedColumnName = $c['referencedColumnName'];
-                $nullable             = !isset($c['nullable']) || $c['nullable'] == true ? 'true' : 'false';
-                $unique               = !isset($c['unique']) || $c['nullable'] == false ? 'false' : 'true';
-                $onDelete             = isset($c['onDelete']) ? '\'' . $c['onDelete'] . '\'' : 'null';
-                $columnDefinition     = isset($c['columnDefinition']) ? '\'' . $c['columnDefinition'] . '\'' : 'null';
+                $nullable = !isset($c['nullable']) || $c['nullable'] == true ? 'true' : 'false';
```
maxbrokman, line 470: Is this behaviour correct? If `$c['nullable']` is not set this will evaluate `true` but isn't doctrine default behaviour to be non null? @patrickbrouwers

maxbrokman:
Looks like you ran the style fixer with the wrong style fixer Matt :p

FoxxMD:
> Looks like you ran the style fixer with the wrong style fixer Matt :p

Shoot really? I thought I used the one in the project. Do I need to specify it as an option in the command?

src/Console/Exporters/FluentExporter.php
FoxxMD: I'm not sure. But the output doesn't even follow that. It outputs each file with `namespace App\Mapping` and originally the filename was the full namespace of the entity being mapped (IE `App.Models.User.php`). That didn't jive well with PSR-4/0 so I changed it to have the filename be the same as the class name for the mapping.

maxbrokman:
Yeah looking at the `'string'.'concates'` are against style and seem to have been autochanged.

I usually just run `php-cs-fixer fix` from the project root and it works

src/Console/Exporters/FluentExporter.php
patrickbrouwers: Default is non null indeed.

src/Console/Exporters/FluentExporter.php
```diff
@@ -12,6 +13,8 @@ class FluentExporter extends AbstractExporter
      */
     protected $_extension = '.php';
 
+    protected $_mappingClassName;
```
patrickbrouwers, line 16: Docblock

src/Console/Exporters/FluentExporter.php
```diff
@@ -22,7 +25,7 @@ class FluentExporter extends AbstractExporter
      */
     public function exportClassMetadata(ClassMetadataInfo $metadata)
     {
-        $lines   = [];
+        $lines = [];
```
patrickbrouwers, line 28: = alignment

src/Console/Exporters/FluentExporter.php
```diff
@@ -32,8 +35,9 @@ public function exportClassMetadata(ClassMetadataInfo $metadata)
             $lines[] = $import;
         }
 
+        $this->_mappingClassName = $this->getClassNameShort($metadata).'Mapping';
         $lines[] = null;
-        $lines[] = 'class ' . $this->getClassNameShort($metadata) . 'Mapping extends ' . $this->getExtendShort($metadata) . ' {';
+        $lines[] = 'class '.$this->_mappingClassName.' extends '.$this->getExtendShort($metadata).' {';
```
patrickbrouwers, line 40: CS spaces

src/Console/Exporters/FluentExporter.php
```diff
@@ -32,8 +35,9 @@ public function exportClassMetadata(ClassMetadataInfo $metadata)
             $lines[] = $import;
         }
 
+        $this->_mappingClassName = $this->getClassNameShort($metadata).'Mapping';
```
patrickbrouwers, line 38: I'd prefer to have this as a getter method instead of property assignment.

src/Console/Exporters/FluentExporter.php
```diff
@@ -51,6 +55,29 @@ public function exportClassMetadata(ClassMetadataInfo $metadata)
         return implode("\n", $lines);
     }
 
+    public function export()
```
patrickbrouwers, line 58: Docblock

patrickbrouwers (changes requested):
Could you run the correct cs-fixer, please. Would make it easier to review the actual changes.

patrickbrouwers:
@FoxxMD could you have a look at my comments?

src/Console/Exporters/FluentExporter.php
guiwoda: I use `App\Infrastructure\Mappings\{entity_name}Mapping.php` in my projects. Not a fan of suffixes for domain classes, but I keep using them in non-domain space.

FoxxMD:
Sorry everyone, I've been out over winter holidays and just had surgery 😬  I will be able to get this cleaned up next week!

PR #195, opened 2016-11-14
[Fix] Correcting a ConvertMapping CLI alias

dragoonis:

The doctrine command syntax for this is:
```
orm:convert-mapping Convert mapping information between supported formats.
```

This is for "converting mapping", whereas you've named this command "convert config" which isn't quite right.

Symfony has correctly named their CLI command: `doctrine:mapping:convert`

I've added an alias to keep your package inline with the rest of the eco-system.

### DIFF
--- a/src/Console/ConvertConfigCommand.php
+++ b/src/Console/ConvertConfigCommand.php
@@ -34,7 +34,7 @@ class ConvertConfigCommand extends SymfonyCommand
     protected function configure()
     {
         $this->setName('doctrine:config:convert')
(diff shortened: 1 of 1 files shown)


patrickbrouwers:
Sorry, but this command has nothing to do with converting mappings?

patrickbrouwers:
I suppose you mean this command https://github.com/laravel-doctrine/orm/blob/1.2/src/Console/ConvertMappingCommand.php, which has the correct signature.

patrickbrouwers:
@dragoonis can you please confirm?

PR #198, opened 2016-12-14
[FIX] Missing state() function on Factory

jakub-klapka:
With laravel/framework 5.3.17, they have added states to model factories. Since function state() is called on every factory() (and entity()) calls, testing with Doctrines Factory will fail on undefined function.

This PR only supresses the error, but does not implement states functionality. It's up to you, if it should be merged.
Ideally, Laravel-Doctrine might adopt states functionality, since it is same, as defineAs functionality. It would be just more familiar to Laravel users. States have one advantage compared to defineAs - you can stack states up.

What approach do you think would be best for Laravel-Doctrine?

### DIFF
--- a/src/Testing/Factory.php
+++ b/src/Testing/Factory.php
@@ -237,4 +237,13 @@ class Factory implements ArrayAccess
     {
         unset($this->definitions[$offset]);
     }
+
+	/**
+	 * Temporary fix to prevent Fatal on Laravel 5.3.17
+	 *
+	 * @return $this
+	 */
+	public function state() {
+		return $this;
+    }
 }
--- a/src/Testing/Factory.php
+++ b/src/Testing/Factory.php
@@ -238,12 +238,13 @@ class Factory implements ArrayAccess
         unset($this->definitions[$offset]);
     }
 
-	/**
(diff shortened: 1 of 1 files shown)


patrickbrouwers:
Hey @jakub-klapka, thanks for your PR.
Could you fix the CS please. :)

I don't know what `state()` does, do you have a link to the documentation about it? But if we can implement it, I'd say it's a good idea.

jakub-klapka:
States are described here: https://laravel.com/docs/5.3/database-testing#writing-factories
The main difference from yours defineAs is, that with states, you define only those properties, which are different from those in base factory. You can define multiple states. And when you are calling factory()->states(['one','two']), it would merge all state properties on top of base factory properties.

I think, that this is really neat, since you can combine those in many different ways. With defineAs, you don't have that much flexibility.

Thanks for considering this ;-)

patrickbrouwers:
Could you open an issue for this, so we can keep this (state support) in mind for the next release?

patrickbrouwers (approved):

jakub-klapka:
Sure, thank you for the merge.

PR #200, opened 2016-12-19
[FEATURE] Alternative Pagination traits

guiwoda:
Alternative Pagination traits to (sort of) decouple from Laravel's Request.
Two Traits were introduced: `PaginatesFromRequest` and `PaginatesFromParams`. The previous `Paginatable` trait was kept for backwards compatilibility, but `PaginatesFromRequest` should be favored.

TO DO:
~~See if https://github.com/doctrine/doctrine2/pull/6180 passes so we can actually unit test this.~~

### DIFF
--- a/src/Pagination/Paginatable.php
+++ b/src/Pagination/Paginatable.php
@@ -2,48 +2,13 @@
 
 namespace LaravelDoctrine\ORM\Pagination;
 
-use Doctrine\ORM\Query;
-
+/**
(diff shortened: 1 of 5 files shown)


guiwoda:
@patrickbrouwers Do you think we could integrate this _as is_, then somewhere along the road add database integration tests for it?

patrickbrouwers (approved):

zdnk:
will you update the documentation?

Issue #187, opened 2016-10-16
[PROPOSAL] Add router extension for implicitly binding doctrine entities


Evertt:
First of all, I would have loved to discuss this first on Slack, but I can't access [the link](http://slack.laraveldoctrine.org/).

Anyhow, since Laravel 5.3, the framework implicitly binds all Eloquent models to routes, if the name of the route-parameter and the variable match up. For for example like this:

``` php
Route::get('/api/users/{user}', function (\App\User $user) {
    // $user is now the correct User object fetched from the DB
});
```

It would be possible to build a router extension which does the same for Doctrine entities. And that would be very nice, since it would remove that one line of unnecessary boilerplate code. :-)

maxbrokman:
If LD already supports model binding I don't see why not. 

We'll have to get the slack link fixed!

patrickbrouwers:
Guess we can support it. Not really a high priority for me though, so PR would be appreciated.

Evertt:
Just to give you a bit of an update. I've been able to fulfil both of my own requests about the router. The code is far from pretty, but at least it works. I wonder if you guys like the code in principle, because if you do then I'm willing to try to make it look prettier.

Here's what I got so far:

``` php
<?php

namespace App;

use Doctrine\ORM\EntityNotFoundException;
use Illuminate\Database\Eloquent\Model;
use Illuminate\Http\Response;
use Spatie\Fractal\ArraySerializer;

class Router extends \Illuminate\Routing\Router
{
    /**
     * Substitute the implicit Doctrine entity bindings for the route.
     *
     * @param  \Illuminate\Routing\Route  $route
     * @return void
     */
    public function substituteImplicitBindings($route)
    {
        parent::substituteImplicitBindings($route);
        $parameters = $route->parameters();
        $entityNamespaces = config('doctrine.managers.default.namespaces');

        /** @var \ReflectionParameter $parameter */
        foreach ($route->signatureParameters() as $parameter) {
            if (!($class = $parameter->getClass())) continue;
            if ($class->isSubclassOf(Model::class)) continue;

            if (!in_array($class->getNamespaceName(), $entityNamespaces)) continue;

            $id = $parameters[$parameter->name];
            $entity = $this->container['em']->find($class->name, $id);

            if (!$entity && !$parameter->isDefaultValueAvailable()) {
                throw EntityNotFoundException::fromClassNameAndIdentifier($class->name, $id);
            }

            $route->setParameter($parameter->name, $entity);
        }
[...]

PR #207, opened 2017-01-24
[FEATURE] Add the routes substitute bindings middleware for doctrine entities

SwenVanZanten:
#187 

I've added the SubstituteBindings middleware which replaces the laravel middleware.
The entity manager will be fetched from the registry class.

I will try to add tests asap!

- Added SubstituteBindings middleware.
- Added Illuminate/routing 5.2.*|5.4.* to composer.json

### DIFF
--- a/src/Middleware/SubstituteBindings.php
+++ b/src/Middleware/SubstituteBindings.php
@@ -106,6 +106,6 @@ class SubstituteBindings
             return (new ReflectionMethod($class, $method))->getParameters();
         }
 
-        return $parameters = (new ReflectionFunction($uses))->getParameters();
+        return (new ReflectionFunction($uses))->getParameters();
     }
 }
--- a/src/Middleware/SubstituteBindings.php
+++ b/src/Middleware/SubstituteBindings.php
@@ -1,0 +1,110 @@
+<?php
+
+namespace LaravelDoctrine\ORM\Middleware;
+
+use Closure;
+use LaravelDoctrine\ORM\IlluminateRegistry;
+use ReflectionMethod;
+use ReflectionFunction;
+use Doctrine\Common\Persistence\Proxy;
+use Illuminate\Contracts\Routing\Registrar;
+use Illuminate\Routing\Middleware\SubstituteBindings as IllumintateSubstituteBindings;
+
+class SubstituteBindings extends IllumintateSubstituteBindings
+{
+    /**
+     * @var IlluminateRegistry
+     */
+    protected $registry;
+
+    /**
+     * Create a new bindings substitutor.
+     *
+     * @param  Registrar          $router
+     * @param  IlluminateRegistry $registry
+     */
+    public function __construct(Registrar $router, IlluminateRegistry $registry)
+    {
+        parent::__construct($router);
+
+        $this->registry = $registry;
+    }
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @param  \Closure                 $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $this->router->substituteBindings($route = $request->route());
+
+        $this->router->substituteImplicitBindings($route);
+
+        $this->substituteImplicitBindings($route);
+
+        return $next($request);
+    }
+
+    /**
+     * Substitute the implicit Doctrine entity bindings for the route.
+     *
+     * @param  \Illuminate\Routing\Route $route
+     * @return void
+     */
+    protected function substituteImplicitBindings($route)
+    {
+        $parameters = $route->parameters();
+
+        $action = $route->getAction();
+
+        foreach ($this->getParameters($action['uses']) as $parameter) {
+            $class = $parameter->getClass()->getName();
+
+            // Try to find the entity manager for the given class.
+            if (is_null($entityManager = $this->getDoctrineEntityManagerByClass($class))) {
+                continue;
+            }
+
+            if (array_key_exists($parameter->name, $parameters)) {
+                $entity = $entityManager->find($class, $parameters[$parameter->name]);
+
+                $route->setParameter($parameter->name, $entity);
+            }
+        }
+    }
+
+    /**
+     * Reflect the parameters of the method or function of the route.
+     *
+     * @param  string | Closure $uses
+     * @return \ReflectionParameter[]
+     */
+    protected function getParameters($uses)
+    {
+        if (is_string($uses)) {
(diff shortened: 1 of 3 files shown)


src/Middleware/SubstituteBindings.php
```diff
@@ -0,0 +1,110 @@
...
+
+use Closure;
+use Doctrine\Common\Persistence\Proxy;
+use Illuminate\Contracts\Routing\Registrar;
+use Illuminate\Routing\Middleware\SubstituteBindings as IllumintateSubstituteBindings;
+use LaravelDoctrine\ORM\IlluminateRegistry;
+use ReflectionFunction;
+use ReflectionMethod;
+
+class SubstituteBindings extends IllumintateSubstituteBindings
```
guiwoda: @HDsign this is confusing, I don't see any calls to `parent` other than the constructor. Is it really necessary to extend Eloquent's middleware?

SwenVanZanten: Well in the handle method the parent router property is called to add the eloquent substitute bindings. This allows the user to get rid of the default SubstituteBindings middleware. Other than that it is not necessary to extend the original middleware.

I guess we could just have both middleware and just add this middleware to add doctrine support.

src/Middleware/SubstituteBindings.php
```diff
@@ -0,0 +1,110 @@
...
+            return (new ReflectionMethod($class, $method))->getParameters();
+        }
+
+        return $parameters = (new ReflectionFunction($uses))->getParameters();
+    }
+
+    /**
+     * Try to fetch the entity manager of the given class.
+     *
+     * @param  string                                     $class
```
patrickbrouwers: string|object

src/Middleware/SubstituteBindings.php
```diff
@@ -0,0 +1,110 @@
...
+     * @return \Doctrine\Common\Persistence\ObjectManager | null
+     */
+    protected function getDoctrineEntityManagerByClass($class)
+    {
+        if (is_object($class)) {
```
patrickbrouwers: In which case could this be an object?

src/Middleware/SubstituteBindings.php
patrickbrouwers: Don't think extending is needed here. Best to define the router property in this class.

src/Middleware/SubstituteBindings.php
```diff
@@ -0,0 +1,110 @@
...
+        foreach ($this->getParameters($action['uses']) as $parameter) {
+            $class = $parameter->getClass()->getName();
+
+            // Try to find the entity manager for the given class.
+            if (is_null($entityManager = $this->getDoctrineEntityManagerByClass($class))) {
+                continue;
+            }
+
+            if (array_key_exists($parameter->name, $parameters)) {
+                $entity = $entityManager->find($class, $parameters[$parameter->name]);
```
patrickbrouwers: How does Eloquent behave in this case? Do they allow null? Or do they thrown an exception if the entity does not exist?

patrickbrouwers: Seems they have something in place for both situations.
https://github.com/laravel/framework/blob/5.3/src/Illuminate/Routing/Router.php#L805

Can you try and make it behave it the same way as Eloquent does.

src/Middleware/SubstituteBindings.php
```diff
@@ -0,0 +1,110 @@
...
+        return $next($request);
+    }
+
+    /**
+     * Substitute the implicit Doctrine entity bindings for the route.
+     *
+     * @param  \Illuminate\Routing\Route $route
+     * @return void
+     */
+    protected function substituteImplicitBindings($route)
```
patrickbrouwers: Typehint Route $route

src/Middleware/SubstituteBindings.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->substituteImplicitBindings($route);
+
+        return $next($request);
+    }
+
+    /**
+     * Substitute the implicit Doctrine entity bindings for the route.
+     *
+     * @param  \Illuminate\Routing\Route $route
+     * @return void
```
patrickbrouwers: Ommit @return void

src/Middleware/SubstituteBindings.php
```diff
@@ -0,0 +1,110 @@
...
+
+                $route->setParameter($parameter->name, $entity);
+            }
+        }
+    }
+
+    /**
+     * Reflect the parameters of the method or function of the route.
+     *
+     * @param  string | Closure       $uses
```
patrickbrouwers: string|callable

patrickbrouwers (changes requested):

src/Middleware/SubstituteBindings.php
```diff
@@ -0,0 +1,109 @@
...
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @param  \Closure                 $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $this->substituteImplicitBindings($request->route());
+
```
patrickbrouwers, line 51: I think the call to `$this->router->substituteBindings($request->route())` could have stayed.

patrickbrouwers (changes requested):

src/Middleware/SubstituteBindings.php
```diff
@@ -0,0 +1,111 @@
...
+     */
+    protected function getParameters($uses)
+    {
+        if (is_string($uses)) {
+            list($class, $method) = explode('@', $uses);
+
+            return (new ReflectionMethod($class, $method))->getParameters();
+        }
+
+        return $parameters = (new ReflectionFunction($uses))->getParameters();
```
Evertt: Why is the `$parameters = ` here? Surely after this function returns (which is on this exact line) the variable `$parameters` will be destroyed right? So what is its purpose on this line?

SwenVanZanten: Hmm, there's absolutely no reason whatsoever 😄 I'll fix it.

patrickbrouwers:
> I will try to add tests asap!

Any ETA on this?

SwenVanZanten:
I'll start on these tonight, try to finish them this weekend 👍

patrickbrouwers:
Great, thanks!

SwenVanZanten:
Guess I've written them already 😄

patrickbrouwers:
Looking good! Will have a last look at it tomorrow.

patrickbrouwers (approved):

PR #205, opened 2017-01-19
[FEATURE] Add support for multiple cache implementations, separating result, query and metadata caching.

BrandonLegault:
This PR adds the ability to configure a different cache driver for each of metadata, result and query caching through the `doctrine.php` config file `'cache'` array.

### Changes proposed in this pull request:

The current cache array format is as follows:

```
'cache'            => [
    'default'      => env('DOCTRINE_CACHE', 'array'),
    'namespace'    => null,
    'second_level' => false,
]
```

This branch introduces this format:

  ```
    'cache' => [
        'second_level'     => false,
        'default'          => env('DOCTRINE_CACHE', 'array'),
        'namespace'        => null,
        'metadata'         => [
            'type'         => 'file',
            'namespace'    => null,
        ],
        'query'            => [
            'type'         => 'redis',
            'namespace'    => null,
        ],
        'result'           => [
            'type'         => 'array',
            'namespace'    => null,
        ],
    ],
```
The configuration is currently applied in this PR by calling `\Doctrine\ORM\Configuration->set[Result|Query|Metadata]CacheImpl` after reading the values from the config.

The `default` cache array entries remain the same to ensure backwards compatibility. If any of the metadata, result or query entries are missing from the configuration then they will take whichever the default entry is. The default entry is also mirrored/stored directly in code in case it's accidentally deleted from the configuration.

PR #211, opened 2017-02-17
[FIX] Fixing default caching bug introduced from #205

ragboyjr:
- Added default cache parameter in createConfiguration

If you have memcached or redis extensions installed
doctrine will try to automatically connect to one of them
if no cache paramter is set when dev mode is off. Defaulting
to ArrayCache to prevent that behavior from occuring.

Signed-off-by: RJ Garcia <rj@bighead.net>

### DIFF
--- a/src/EntityManagerFactory.php
+++ b/src/EntityManagerFactory.php
@@ -2,6 +2,7 @@
 
 namespace LaravelDoctrine\ORM;
 
+use Doctrine\Common\Cache\ArrayCache;
 use Doctrine\Common\Cache\Cache;
 use Doctrine\ORM\Cache\DefaultCacheFactory;
 use Doctrine\ORM\Configuration;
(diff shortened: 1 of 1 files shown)


src/EntityManagerFactory.php
```diff
@@ -94,7 +95,8 @@ public function create(array $settings = [])
     {
         $configuration = $this->setup->createConfiguration(
             array_get($settings, 'dev', false),
-            array_get($settings, 'proxies.path')
+            array_get($settings, 'proxies.path'),
+            new ArrayCache()
```
patrickbrouwers: It's probably best to get the actual default cache driver then:

$defaultDriver    = $this->config->get('doctrine.cache.default', 'array');
$this->cache->driver($driver);

patrickbrouwers (changes requested):

ragboyjr:
@patrickbrouwers no problem, can you check now?

nghiaminhle:
i have the same problem. Please add this line into EntityFactoryManager:
 
    public function create(array $settings = [])
    {
        $configuration = $this->setup->createConfiguration(
            array_get($settings, 'dev', false),
            array_get($settings, 'proxies.path'),
            **$this->cache->driver()**
        );
----------

Issue #202, opened 2016-12-29
[BUG] Lumen overwrites validation.presence with Illuminate\Validation\DatabasePresenceVerifier


blacktemplar:
I am trying to use the presence validation for doctrine, but I cannot use it with Lumen since it gets overwritten by the default presence provider. I am using the Lumen Framework 5.3.3.

I debugged it a little bit and the problem seems to be that registerValidatorBindings() in Laravel\Lumen\Application.php binds a function which calls register('Illuminate\Validation\ValidationServiceProvider'). In the ValidationServiceProvider the validation.presence singleton gets overwritten. So it seams to be a problem with the Lumen Framework.

patrickbrouwers:
Not really sure that's something we can fix in the library to be honest :(
You could try and overrule that `registerValidatorBindings` in your project and set it to the doctrine service provider.

matt-allan:
This is a general problem with how Lumen works ([see this issue](https://github.com/laravel/lumen-framework/issues/416)).

@blacktemplar If you resolve the validator once (`app('validator')`) before the Laravel Doctrine service provider runs it should work.  That will stop [this block of code](https://github.com/laravel/lumen-framework/blob/5.2/src/Application.php#L199) from running.

patrickbrouwers:
Should be fixed in 1.3.4

PR #215, opened 2017-03-09
[FIX] Presence verifier on lumen

Lidbetter:
- Unbind default framework binding implementations of 'validator' (if wanting to use doctrine presence)
- Tweak binding of  `PresenceVerifierProvider` to account for the removal of the default  'validator' binding (if wanting to use doctrine presence)
- Ensure the default implementation of `ValidationServiceProvider` has a db instance to use when checking to see if it should set the presence verifier (when using the default presence)

### DIFF
--- a/src/Validation/PresenceVerifierProvider.php
+++ b/src/Validation/PresenceVerifierProvider.php
@@ -2,6 +2,7 @@
 
 namespace LaravelDoctrine\ORM\Validation;
 
+use Illuminate\Validation\Factory;
 use Illuminate\Validation\ValidationServiceProvider;
 
 class PresenceVerifierProvider extends ValidationServiceProvider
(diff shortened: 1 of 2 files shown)


Lidbetter:
Fixes #202

rgwsk65:
Is there a chance of merge this PR?

src/Validation/PresenceVerifierProvider.php
```diff
@@ -14,6 +15,27 @@ class PresenceVerifierProvider extends ValidationServiceProvider
     protected $defer = true;
 
     /**
+     * Register the validation factory.
+     *
+     * @return void
+     */
+    protected function registerValidationFactory()
```
patrickbrouwers, line 22: Is it necessary to redeclare this if we are extending `ValidationServiceProvider` ?

patrickbrouwers (changes requested):

src/Validation/PresenceVerifierProvider.php
Lidbetter: Yes, the default implementation checks for different unrelated conditions before calling `$validator->setPresenceVerifier`

patrickbrouwers (approved):

PR #224, opened 2017-04-03
[FEATURE] Add support for default table options in configuration

hultberg:
### Changes proposed in this pull request:
- Supporting `defaultTaleOptions` in configration resolving. This is supported by doctrine and enables developers to set a default collation on all tables.

I ran into some troubles with the php-cs-fixer config in this project as I was using php-cs-fixer 2.0.. However, I manually ran cs fixer on the files I modified.

I also noticed that StyleCI conflicts with what php cs fixer.. fixes.

PR #225, opened 2017-04-03
[FEATURE] Add support for default table options in configuration

hultberg:
(Original: #224)

- Supporting `defaultTableOptions` in configration resolving. This is supported by doctrine and enables developers to set a default collation on all tables.

I was unable to make all tests work, as they seem to have broken on 1.2.


patrickbrouwers:
Released for both 1.2 as 1.3.

PR #228, opened 2017-04-12
[FEATURE] Add ability to provide different folders for Entity Factories via doctrine.php config

Shandur:
- Using doctrine.php config file user can use his own Entity Factory folders. 

_Reason:_
I was stuck with a problem: I'm developing an application using modular approach. Each module has its own entities and its own factories for them. But I couldn't run the factories because of hard-coded folder that leads to 'database/factories' in the project root.

Thanks!

### DIFF
--- a/config/doctrine.php
+++ b/config/doctrine.php
@@ -204,5 +204,21 @@ return [
      */
     'notifications'              => [
         'channel' => 'database'
+    ],
+
+    /*
+     |--------------------------------------------------------------------------
+     | Doctrine Entity Factories
+     |--------------------------------------------------------------------------
+     |
+     |  You can add custom folders with your factories, for instance, if you have
+     |  modular system and each module contains its own factories.
+     |  If no folders are specify, the default one will be used ('database/factories' in project root)
+     |
+     */
+    'factories' => [
+        'folders' => [
+            'Modules/User/Database/Factory'
+        ]
     ]
 ];
--- a/config/doctrine.php
+++ b/config/doctrine.php
@@ -218,7 +218,8 @@ return [
      */
     'factories' => [
         'folders' => [
-            'Modules/User/Database/Factory'
+            // 'Modules/User/Database/Factory',
+            // 'Modules/Features/Database/Factory'
         ]
     ]
 ];
--- a/config/doctrine.php
+++ b/config/doctrine.php
@@ -218,8 +218,8 @@ return [
      */
     'factories' => [
         'folders' => [
-            // 'Modules/User/Database/Factory',
-            // 'Modules/Features/Database/Factory'
+            database_path('factories'),
+            // base_path('Modules/User/Database/Factory'),
         ]
     ]
 ];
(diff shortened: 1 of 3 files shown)


config/doctrine.php
```diff
@@ -204,5 +204,22 @@
...
+     |--------------------------------------------------------------------------
+     |
+     |  You can add custom folders with your factories, for instance, if you have
+     |  modular system and each module contains its own factories.
+     |  If no folders are specify, the default one will be used ('database/factories' in project root)
+     |
+     */
+    'factories' => [
+        'folders' => [
+            // 'Modules/User/Database/Factory',
```
patrickbrouwers: Can you add `database_path('factories')` as default here.
Also probably better to set the example to `base_path('Modules/...')` ?

src/Testing/Factory.php
```diff
@@ -55,9 +55,11 @@ public static function construct(Faker $faker, ManagerRegistry $registry, $pathT
 
         $factory = new static($faker, $registry);
 
-        if (is_dir($pathToFactories)) {
-            foreach (Finder::create()->files()->in($pathToFactories) as $file) {
-                require $file->getRealPath();
+        foreach ((array)$pathToFactories as $factoryDir) {
```
patrickbrouwers: Code style: space behind `(array)`

patrickbrouwers (changes requested):
Nice idea :)

guiwoda:
My 2c: Don't use scripts if you want to modularize. Use ServiceProviders.

Sample:

```php
use LaravelDoctrine\ORM\Testing\Factory as EntityFactory;

class SomeModuleServiceProvider extends ServiceProvider
{
    public function register()
    {
        $this->app->afterResolving(EntityFactory::class, function (EntityFactory $factory) {
            $factory->define(SomeEntity::class, function ($faker) {
                // your definition here :-)
            });
        });
    }
}
```

patrickbrouwers:
Actually have to agree with @guiwoda. Sounds indeed a better idea to modularize :)

Shandur:
@guiwoda @patrickbrouwers Probably you're right about using service providers for adding entity factories. But if you have about 10 or more entities and you need factories for them, it's not comfortable inject them into service provider. 
I guess it could be better to add entity folders to doctrine config via service provider dynamically for each module (while register) instead of writing them directly into the file when you're using modular approach :)

config/doctrine.php
Shandur: @patrickbrouwers sure, thanks :) it looks better

guiwoda:
Then do something like...

```php
use LaravelDoctrine\ORM\Testing\Factory as EntityFactory;

class SomeModuleServiceProvider extends ServiceProvider
{
    public function register()
    {
        $this->app->afterResolving(EntityFactory::class, function (EntityFactory $factory) {
            foreach ($this->getFactoryFolders() as $path) {
                require $path;
            }
        });
    }
}
```

guiwoda:
I'm sorry but your use case is not something I'd want to generalize in the package, and if it were, I don't agree with what you're proposing either. I appreciate your intention to contribute, but I have to say I don't think this needs to be merged.

Shandur:
@guiwoda Ok, it's up to you, thanks for you reply :)

patrickbrouwers:
Closing this in favour of registering it through the service provider

PR #235, opened 2017-05-11
[FEATURE] Show Entity class name when no factory exists

Korri:
Mimics Laravel's original way of doing it and helps find what needs to be implemented.

- Will print class name after factory name when no factory is found.

This:
```
Unable to locate factory with name [default].
```
Becomes this:
```
Unable to locate factory with name [default] [App\User].
```


patrickbrouwers:
Thanks for your contribution!

PR #236, opened 2017-05-25
[FIX] Don't use  for all as there is no join

Korri:
- Set `$fetchJoinCollection` to false for `paginateAll` queries as we are sure that they won't do any join, this will reduce the number of queries by 1.


patrickbrouwers:
Thx 👍

Issue #238, opened 2017-06-06
[BUG?] SimpleHydrator should be calling setter method if available


Fluepke:
In Doctrine it's the application's responsibility of keeping a bidirectional association in sync.

In their [documentation](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/unitofwork-associations.html#important-concepts) it reads as follows:

> Make sure to update both sides of a bidirectional association (or at least the owning side, from Doctrine’s point of view).

When creating an object using a `Factory`, _laravel-doctrine_ relies on the `SimpleHydrator` to create the actual object from the array provided within the custom `Factory`.

The `SimpleHydrator` however does not call the **setter methods** which may be used to keep the _inverse side_ in sync, instead it relies on setting the properties directly.

Furthermore setters might perform checks and validation, so IMO it's reasonable to first check if a setter method exists and if not fallback to setting the property directly.

guiwoda:
Closing this, check #239 for more details.

PR #239, opened 2017-06-06
[Fix] #238 by calling the setter method if available.

Fluepke:
This is a fix for #238 by simply calling the setter if there is one.

### DIFF
--- a/src/Testing/SimpleHydrator.php
+++ b/src/Testing/SimpleHydrator.php
@@ -32,7 +32,11 @@ class SimpleHydrator
      */
     private static function hydrateReflection(ReflectionClass $reflection, $instance, $field, $value)
     {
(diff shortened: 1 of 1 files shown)


guiwoda:
Hi @Fluepke!
I believe you may be misusing entity factories: They are not meant for production code.

Entity factories are helpers for testing and database seeding. Production code should never use reflection and property injection like this, ever. You should always model your constructor(s) and call on them wherever you need to instantiate an Entity.

This PR suggests that Entities should have setter methods on them, which I also don't agree with as a recommended design decision.

I'll be closing this and #238, please join us in Slack if you want to discuss this further!

Fluepke:
Ok. This was only intended for testing of course.
IMO setters are an easy way of keeping the inverse side updated, but I'll join your discussion on Slack.

PR #240, opened 2017-06-07
[FIX] Convert seconds to milliseconds

Korri:
- See [comment](https://github.com/doctrine/dbal/pull/2739#issuecomment-306880407) `executionMS` is actually seconds not milliseconds.


patrickbrouwers:
Hey @Korri, thanks for the contribution, can you also please fix the test, it's failing now.

Korri:
Here you go!

Issue #237, opened 2017-05-31
[PROPOSAL] Make DoctrineCollector widget keys configurable or unique


zquintana:
Right now in the current model the collector for debugbar collides widget keys with Laravel's built in query collector. This makes it difficult to use Doctrine along side Eloquent with debugging for monitoring. With existing applications it is desirable at times to keep eloquent running alongside Doctrine during the transition. Allowing this to be configurable or using a different key allows both logs to exist side by side.

see https://github.com/laravel-doctrine/orm/blob/1.3/src/Loggers/Debugbar/DoctrineCollector.php#L26

If this is need by more people than myself, I can fork this repo and contribute it.

guiwoda:
I think reading an env variable and defaulting to the current key would be fine.
Your PR would be welcome, @zquintana

patrickbrouwers:
Closing this one as we have a PR #241 for it.

PR #241, opened 2017-06-12
[FEATURE] Adds debugbar key to doctrine collector.

zquintana:
This PR is for the proposal in https://github.com/laravel-doctrine/orm/issues/237. It adds config for the doctrine debugbar widget name. It was mentioned to use the `env`, but I added it to the configs because the use of `env` outside of configs is deprecated in laravel 5.2+ and it's use doesn't work when configs are cached.

The new config name is `debugbar_widget_key` and it defaults to the current value. Setting it in config will allow the eloquent debugbar to show with the doctrine debugbar.


src/Loggers/Debugbar/DoctrineCollector.php
```diff
@@ -13,6 +14,23 @@ class DoctrineCollector extends DebugbarDoctrineCollector
     protected $debugStack;
 
     /**
+     * @var string
+     */
+    protected $name;
```
guiwoda: I'd stay consistent with the `key` keyword here as well.

guiwoda (changes requested):

guiwoda:
Please fix the tests as well

config/doctrine.php
```diff
@@ -143,6 +143,8 @@
     |--------------------------------------------------------------------------
     */
     'logger'                     => env('DOCTRINE_LOGGER', false),
+    'debugbar_widget_key'        => env('DOCTRINE_DEBUGBAR_KEY', 'queries'),
```
patrickbrouwers: Don't think it's really needed to have a env variable for this. You would change this setting for your project, not per environment.

src/Loggers/Debugbar/DoctrineCollector.php
```diff
@@ -13,6 +14,23 @@ class DoctrineCollector extends DebugbarDoctrineCollector
...
+
+
+    /**
+     * DoctrineCollector constructor.
```
patrickbrouwers: Ommit description

src/Loggers/Debugbar/DoctrineCollector.php
patrickbrouwers: I would prefer `widgetName` as that would be more explicit.

config/doctrine.php
patrickbrouwers: Also think this should be `debugbar_widget_name`.

patrickbrouwers (changes requested):
Thanks for your PR, looking good; just a few small remarks. Also please fix the failing tests.

config/doctrine.php
zquintana: Changed it to `debugbar_widget_name` in latest commit.

src/Loggers/Debugbar/DoctrineCollector.php
zquintana: Removed the description too.

patrickbrouwers (changes requested):
Please fix tests + code-style issue. (See merge checks)

patrickbrouwers:
Replaced

PR #242, opened 2017-06-12
[FIX] Throw an exception if exist: rule refers to an invalid entity type

Korri:
### Current behavior
When using the `exists:` rule with `DoctrinePresenceVerifier` on a non-mapped class, the following error is thrown: `Call to a member function createQueryBuilder() on null`
### Proposed behavior
Throw an easy to understand `InvalidArgumentException`: `\App\User is not a valid Doctrine entity type`

### DIFF
--- a/src/Validation/DoctrinePresenceVerifier.php
+++ b/src/Validation/DoctrinePresenceVerifier.php
@@ -5,6 +5,7 @@ namespace LaravelDoctrine\ORM\Validation;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\QueryBuilder;
 use Illuminate\Validation\PresenceVerifierInterface;
+use InvalidArgumentException;
 
 class DoctrinePresenceVerifier implements PresenceVerifierInterface
 {
@@ -90,9 +91,6 @@ class DoctrinePresenceVerifier implements PresenceVerifierInterface
     protected function select($collection)
     {
         $em = $this->getEntityManager($collection);
-        if ($em === null) {
-            throw new \InvalidArgumentException("[$collection] is not a valid Doctrine type.");
-        }
 
         $builder = $em->createQueryBuilder();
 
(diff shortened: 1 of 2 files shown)


guiwoda (approved):

guiwoda (changes requested):
Missing test reflecting change?

Korri:
Here you go

src/Validation/DoctrinePresenceVerifier.php
```diff
@@ -89,7 +89,11 @@ public function getMultiCount($collection, $column, array $values, array $extra
      */
     protected function select($collection)
     {
-        $em      = $this->getEntityManager($collection);
+        $em = $this->getEntityManager($collection);
+        if ($em === null) {
+            throw new \InvalidArgumentException("[$collection] is not a valid Doctrine type.");
```
patrickbrouwers: Can you move this if-statement to the `getEntityManager` method. It should only have to be implemented for `$this->registry->getManagerForClass($entity)` .

Change message to `No Entity Manager could be found for [%s]`.

Also import class + use sprintf.

tests/Validation/DoctrinePresenceVerifierTest.php
```diff
@@ -123,6 +123,17 @@ public function test_can_get_multi_count_with_extra_conditions()
         ]);
     }
 
+    public function test_counting_invalid_entity_throws_exception()
+    {
+        $this->registry->shouldReceive('getManagerForClass')
+                       ->with(CountableEntityMock::class)
+                       ->andReturn(null);
+
+        $this->expectException(InvalidArgumentException::class);
```
patrickbrouwers: Also assert the exception message.

patrickbrouwers (changes requested):
Thanks for your PR, nice change. Can you please have a look at my CR comments.

Korri:
Should look good now !

patrickbrouwers (approved):
Looks good! What do you think @guiwoda ?

guiwoda (approved):

guiwoda:
Looks good to me too!

PR #249, opened 2017-06-28
[FIX] Add return statement to Illuminate::resetManager()

adridev:
- Add return statement to Illuminate::resetManager()
- Update related tests to check that behavior


patrickbrouwers:
Will have to look into the boot extensions my self. Perhaps good to remove them from the PR, so I can merge the return statement?

adridev:
Close this. I will open a new one with only the return in resetManager method.

PR #255, opened 2017-07-10
[FIX] Use container to create ValidationFactory

Korri:
- Use `Application::make` instead of `new` to create `Validation\Factory` instance


Korri:
Never mind, this breaks my code, looks like this is called to early.

Issue #256, opened 2017-07-14
[PROPOSAL] Add ability to register custom hydrators through config


Viniter:
There is currently no nice way of registering custom hydrators, other then fetching the entity manager somewhere in the application bootstrap and calling `$em->getConfiguration()->addCustomHydrationMode()`.

It would be nice if there was an optional field in the configuration, and it would be registered in the `EntityManagerFactory` similarly to custom types or custom functions.

It's a tiny change, I could submit a PR if you're interested.

patrickbrouwers:
PR's are always welcomed :)

PR #257, opened 2017-07-14
[FEATURE] Add hydrator registration to EntityManagerFactory

Viniter:
* Enables an easy way to register custom Doctrine hydrators
* Add `custom_hydration_modes` entry to `doctrine` config
* Add `setCustomHydrationModes` method to EntityManagerFactory
* Implements #256

### DIFF
--- a/src/EntityManagerFactory.php
+++ b/src/EntityManagerFactory.php
@@ -328,7 +328,7 @@ class EntityManagerFactory
      */
     protected function setCustomHydrationModes(Configuration $configuration)
     {
-        $hydratorConfig = $this->config->get('doctrine.custom_hydration_modes');
+        $hydratorConfig = $this->config->get('doctrine.custom_hydration_modes', []);
         foreach ($hydratorConfig as $hydrationModeName => $customHydratorClass) {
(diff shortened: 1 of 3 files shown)


Viniter:
Please verify I modified `EntityManagerFactoryTest` correctly, I'm not used to working with Mockery.

src/EntityManagerFactory.php
```diff
@@ -325,6 +326,17 @@ protected function setCustomFunctions(Configuration $configuration)
     /**
      * @param Configuration $configuration
      */
+    protected function setCustomHydrationModes(Configuration $configuration)
+    {
+        $hydratorConfig = $this->config->get('doctrine.custom_hydration_modes');
```
patrickbrouwers: For backwards compatability it should be `$this->config->get('doctrine.custom_hydration_modes', []);`

If we don't add the default `[]`, it will throw an error that it can't foreach a `null`

patrickbrouwers (changes requested):

src/EntityManagerFactory.php
Viniter: Fixed now.

patrickbrouwers (approved):

patrickbrouwers:
Thx!

PR #265, opened 2017-08-07
[FIX] Remove  @ORM\Entity in Superclass

bigperson:
`@ORM\Entity` сaused an error:
Column not found: 1054 Unknown column 'n0_.id' in 'field list'

- Remove @ORM\Entity in Superclass


bigperson:
Why tests is not successful? It is not relevant to my commit :(

PR #268, opened 2017-08-22
[FEATURE] Testing factory states

guiwoda:
Added support for testing factory states, as in Eloquent.


patrickbrouwers:
Looks good, will play around with it tomorrow

PR #269, opened 2017-08-22
[FIXish] Travis fixes

guiwoda:
- The Cache adapter was using strings on a mock method that was supposed to be returning an int.
- The JsonType is not compatible with doctrine/dbal 2.6. Current test fetches the "json" type from Doctrine, so I assumed it was testing compatibility, not only this implementation. This will break backwards compat tho.

### DIFF
--- a/src/Types/Json.php
+++ b/src/Types/Json.php
@@ -36,7 +36,11 @@ class Json extends JsonArrayType
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
-        return $value === null ? null : parent::convertToPHPValue($value, $platform);
+        if ($value === null || $value === '') {
+            return null;
+        }
+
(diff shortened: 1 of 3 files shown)


src/Types/Json.php
```diff
@@ -36,7 +36,11 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
-        return $value === null ? null : parent::convertToPHPValue($value, $platform);
+        if ($value === null || $value === '') {
+            return null;
```
patrickbrouwers, line 40: I think this was actually a conscious decision and would break apps if we change it.

patrickbrouwers (changes requested):

PR #270, opened 2017-08-30
[FIX] Add Laravel 5.5 support

DuckThom:
- Adding Laravel 5.5.* as a dependency allowing `laravel-doctrine/orm` to be installed on 5.5

Note: I briefly tested it under 5.5 and I did not run into any issues. However, PHPUnit showed some failed tests but they did not appear to be caused by Laravel 5.5 as I got the same failed tests when i checked out the latest stable version (1.3.7) and ran phpunit on there.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,15 +18,15 @@
   "require": {
     "php": ">=5.6.4",
     "doctrine/orm": "2.5.*",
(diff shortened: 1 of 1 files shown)


patrickbrouwers:
Thanks for the PR, I'll take care of the failing tests :)

PR #286, opened 2017-10-18
Added master/slave connection support to branch 1.2.

aledelgo:
[FEATURE].

Hi, i need master/slave feature in my laravel 5.3 project. i cannot move to 5.4 (to use l-d 1.3) and this feature has not yet been merged in laravel-doctrine 1.2

reading commits comments it seems this pull don't have any breaking-change from 1.2 to 1.3 so i think we could test it.


patrickbrouwers:
1.2 is no longer supported. Also due to the failing tests there's no way I can safely merge this.

PR #288, opened 2017-10-26
[FIX] Optimize UserProvider if just one Object Manager

eigan:
I was doing some profiling with xdebug, and noticed that the UserProvider was spending **30%** of the entire request inside `getAllMetadata`. This led me up to this method that will verify that `$class` (our User class) is actually managed by the ObjectManager returned from `getManagerNames`.

Since we only have one EntityManager, I thought that this wasnt necessary, so just doing an early return instead. 

Our testsuite is still OK with this change, but Iam not familiar enough with doctrine to know if this is OK..

### DIFF
--- a/src/IlluminateRegistry.php
+++ b/src/IlluminateRegistry.php
@@ -336,7 +336,13 @@ final class IlluminateRegistry implements ManagerRegistry
             $class = $proxyClass->getParentClass()->getName();
         }
 
-        foreach ($this->getManagerNames() as $name) {
+        $managerNames = $this->getManagerNames();
+
+        if (count($managerNames) === 1) {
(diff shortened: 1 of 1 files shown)


eigan:
If we take a look at the [AbstractManagerRegistry](https://github.com/doctrine/common/blob/master/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php#L189), the only difference between the current implementation is that we execute getAllMetadata(). 

An alternative patch, would be something like this:

```diff
if (!$manager->getMetadataFactory()->isTransient($class)) {
-    foreach ($manager->getMetadataFactory()->getAllMetadata() as $metadata) {
-        if ($metadata->getName() === $class) {
            return $manager;
-        }
-    }
}
```

PR #289, opened 2017-10-26
Fix DefaultQueryCache not knowing that we put anything to the store

eigan:
There is a mismatch between the two cache interfaces:

- Doctrine\Common\Cache\Cache::save: bool
- Illuminate\Contracts\Cache\Repository::put(): void

See the laravel implementation here [Illuminate\Cache\Repository](https://github.com/illuminate/cache/blob/master/Repository.php#L188). (called via $this->cache->put())

I noticed this while stepping into DefaultQueryCache::put() when trying to cache the result of a query. [See the call here](https://github.com/doctrine/doctrine2/blob/07b397f3419e37f4ac0bf1d0a55572af7cb9d575/lib/Doctrine/ORM/Cache/DefaultQueryCache.php#L282).

### DIFF
--- a/src/Configuration/Cache/IlluminateCacheAdapter.php
+++ b/src/Configuration/Cache/IlluminateCacheAdapter.php
@@ -51,7 +51,9 @@ class IlluminateCacheAdapter extends CacheProvider
     protected function doSave($id, $data, $lifeTime = false)
     {
         if (!$lifeTime) {
-            return $this->cache->forever($id, $data);
+            $this->cache->forever($id, $data);
+
+            return true;
         }
 
         // Laravel cache system accept expire times in minutes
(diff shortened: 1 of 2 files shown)


eigan:
Could probably return has() instead

eigan:
Fixed the `$this->forever()` call too, which should fix the issue that `doctrine:clear:metadata:cache` always reports _"No cache entries were deleted."_. See the [call into doSave()](https://github.com/doctrine/cache/blob/beb0fa35b61e9073f8612d9ffd34920bdaec406a/lib/Doctrine/Common/Cache/CacheProvider.php#L176), which reports [back to the command](https://github.com/laravel-doctrine/orm/blob/5d1d09bd8956eb7a33a88cd3dcc567f2ea361605/src/Console/ClearMetadataCacheCommand.php#L54).

eigan:
Note that in L5.5, the set method from PSR have been introduced, which should return a bool, but laravel don't respect that.

guiwoda:
Could you add a comment on the adapter to explain why we hard-code the `return true`?

PR #290, opened 2017-10-29
[FIX] DoctrineUserProvider uses wrong return type

eigan:
Just a documentation error where an invalid interface was used.

- Remove usage of `LaravelDoctrine\ORM\Contracts\Auth\Authenticatable` as an interface (`getEntity`)
  - It is not an interface but a trait
- Renamed `IlluminateAuthenticatable` to just `Authenticatable`
  - Because many other classes was using just `Authenticatable`


eigan:
Sorry, branched out from a previous PR. See #291

PR #294, opened 2017-11-16
[FIX] Stop execution of command after successful drop

eigan:
This fixes issue where you would get output like this:

```
[..]
Database schema dropped successfully!
Nothing to drop. The database is empty!
```


guiwoda:
I think this command deserves a refactor. I'd rather clean those ifs up instead of adding an early return

PR #295, opened 2017-11-30
[FIX] Use cache for annotation reader

R3VoLuT1OneR:
Annotation reader using `ArrayCache` that is performance killer.
http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html#setup-and-configuration

- Implemented usage of default cache for annotation reader.

### DIFF
--- a/src/Configuration/MetaData/Annotations.php
+++ b/src/Configuration/MetaData/Annotations.php
@@ -2,10 +2,21 @@
 
 namespace LaravelDoctrine\ORM\Configuration\MetaData;
 
-use Doctrine\ORM\Configuration;
+use Doctrine\Common\Annotations\AnnotationReader;
+use Doctrine\Common\Annotations\AnnotationRegistry;
+use Doctrine\Common\Annotations\CachedReader;
+use Doctrine\Common\Annotations\SimpleAnnotationReader;
+use Doctrine\Common\Cache\ArrayCache;
+use Doctrine\Common\Cache\Cache;
+use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
 
 class Annotations extends MetaData
 {
+    /**
+     * @var Cache
+     */
+    protected $cache;
+
     /**
      * @param array $settings
      *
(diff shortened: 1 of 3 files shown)


patrickbrouwers:
I'm wondering if this will actually change performance. The annotation reader only gets a cached reader if cache is enabled. If cache is enabled the entire metadata is cached, so it wouldn't even reach the annotation reader?

R3VoLuT1OneR:
@patrickbrouwers Actually it is not increased performance for me, after more tests and I disabled this feature. You right, if metacache enabled it should not reach the cache reader cache, but it is not the case, it is do go any request.

So this MR need much more research, maybe close it.

patrickbrouwers:
If you find out anything, let us know!

patrickbrouwers:
Reopening because of #320

bjdelange:
Thank you @R3VoLuT1OneR for your work.

This PR fixes the problem perfectly. I have confirmed through xdebug profiling that annotation parsing no longer happens and everything comes from cache. The performance improvement is huge, easily visible even with the naked eye ( I have quite a lot of entity classes so that helps ;-) )

bjdelange (approved):

R3VoLuT1OneR:
@bjdelange Good! I seen it by documentation and code reading but it is not helped with our issues, so leaved it without even testing =\

bicatu:
what are you talking about?

On Wed, Mar 21, 2018 at 3:06 PM, Pavel Zhytomirsky <notifications@github.com

bjdelange:
@R3VoLuT1OneR what do you mean you left it without testing ?
I copied your code without making *any* changes and it works perfectly.

bjdelange:
TL;DR a recent commit to the 1.4 branch breaks this fix

To test this fix some more I updated my fork, applied these changes and loaded the forked package (https://github.com/bjdelange/orm/tree/annotation-cache) into my project. 
Now, after clearing the cache, I'm running into max nesting level problems.
When I revert back to the previous package version, https://github.com/bjdelange/orm/tree/master-slave-wrapperclass, everything is back to normal.
The setup I used yesterday was my master-slave-wrapperclass fork with this patch applied, so I did that and it works perfectly again.

bjdelange (changes requested):
I don't know why but applying this patch causes a max_nesting_level problem on the latest version of this branch. On a slightly older version, just a few commits ago, it works fine.

realshadow:
@bjdelange I am currently on the 1.4.4 tag, because update to doctrine 2.6 breaks few things. I can confirm the PR works, but I have hard time spotting any difference in load times (PHP 7.2, opcache, redis cache). What does it do to load times for you?

bjdelange:
It improves load times by more than 40% for me.

bjdelange:
The max nesting level problem has been fixed: https://github.com/laravel-doctrine/extensions/commit/a2e7896100559ecc64504252dbc74ec18ad59813
(I just discovered it was already fixed; was working on it myself and came up with the exact same fix)

porozhnyy:
I checked with myself - this code really speeds up the application. Is there any reason not to apply it?

eigan:
This is not needed for me if I enable metadata cache. _(I am not using any extensions)._

#### Steps to verify:
1) Enable a cache driver for metadata (I used redis)
2) Set xdebug breakpoing on `vendor/doctrine/persistence/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php` line `178`.
3) Notice the metadata is retrieved from cache.

See the AbstractClassMetadataFactory here on github: https://github.com/doctrine/persistence/blob/af1ec238659a83e320f03e0e454e200f689b4b97/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php#L177

eigan:
I would be greatful if anyone could give me step-by-step instructions to reproduce this problem.

The `AnnotationDriver::reader` is only used when there is no cached metadata. So using the already existing config for that (doctrine.cache.metadata.driver) should be sufficient (as @patrickbrouwers already points out).

_EDIT: Closing because of no activity for the past year. If this was a problem for more people, then we should have seen a lot more activity here._

PR #301, opened 2018-01-29
[FIX] Depend on AbstractQuery instead of Query

bjdelange:
- fix two method parameters to require AbstractQuery instead of Query (which is final) so it can be used with mocks for testing

### DIFF
--- a/src/Pagination/PaginatesFromParams.php
+++ b/src/Pagination/PaginatesFromParams.php
@@ -2,7 +2,7 @@
 
 namespace LaravelDoctrine\ORM\Pagination;
 
-use Doctrine\ORM\Query;
+use Doctrine\ORM\AbstractQuery;
 
 trait PaginatesFromParams
 {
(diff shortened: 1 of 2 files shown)


src/Pagination/PaginatesFromParams.php
```diff
@@ -20,14 +20,14 @@ public function paginateAll($perPage = 15, $page = 1)
...
-     * @param bool  $fetchJoinCollection
+     * @param AbstractQuery $query
+     * @param int           $perPage
+     * @param int           $page
+     * @param bool          $fetchJoinCollection
      *
      * @return \Illuminate\Pagination\LengthAwarePaginator
      */
-    public function paginate(Query $query, $perPage, $page = 1, $fetchJoinCollection = true)
+    public function paginate(AbstractQuery $query, $perPage, $page = 1, $fetchJoinCollection = true)
```
patrickbrouwers, line 30: AbstractQuery is not imported. Did you test this?

patrickbrouwers (changes requested):

src/Pagination/PaginatesFromParams.php
bjdelange: Oops.. copy paste mistake. I found this one just as I was reviewing so it was a last minute change.

patrickbrouwers (approved):

PR #307, opened 2018-02-12
[FEATURE] Allow Lumen Publishing

ragboyjr:

- Allows for publishing config files in lumen.

Even though lumen doesn't provide vendor publishing by default, it's easily enabled by just adding in the `Illuminate\Foundation\Console\VendorPublishCommand`. However, since you have the `isLumen` check, no laravel-doctrine files get published.

### DIFF
--- a/src/DoctrineServiceProvider.php
+++ b/src/DoctrineServiceProvider.php
@@ -49,11 +49,9 @@ class DoctrineServiceProvider extends ServiceProvider
         $this->extendAuthManager();
         $this->extendNotificationChannel();
 
-        if (!$this->isLumen()) {
-            $this->publishes([
-                $this->getConfigPath() => config_path('doctrine.php'),
-            ], 'config');
-        }
+        $this->publishes([
+            $this->getConfigPath() => config_path('doctrine.php'),
+        ], 'config');
 
         $this->ensureValidatorIsUsable();
     }


ragboyjr:
ping

ragboyjr:
ping

src/DoctrineServiceProvider.php
```diff
@@ -49,11 +49,9 @@ public function boot()
         $this->extendAuthManager();
         $this->extendNotificationChannel();
 
-        if (!$this->isLumen()) {
-            $this->publishes([
-                $this->getConfigPath() => config_path('doctrine.php'),
-            ], 'config');
-        }
+        $this->publishes([
+            $this->getConfigPath() => config_path('doctrine.php'),
```
patrickbrouwers, line 53: Clean lumen install doesn't have a config_path helper.

patrickbrouwers (changes requested):

ragboyjr:
@patrickbrouwers didn't mean to annoy you. I get a flood of emails every day it feels like, sometimes, I see a Github notification but then forget about it that day.

I'm closing this because the laravel foundation/helpers conflict with the lumen helpers. When I first tried this, the lumen helpers loaded first, but when used composer install from a fresh install, laravel's went first and it broke the system. So unfortunately you can't have laravel/framework installed with lumen :(.

Thank you by the way for taking the time to make this package and review the code. It's been very helpful in my projects.

PR #308, opened 2018-02-14
[FIX/FEATURE] Don't extend the auth manager unless it's been bound

### LINKED ISSUE #302: [PROPOSAL] Don't extend the auth manager unless it's been bound

1.4, 5.5

### Expected behaviour

https://github.com/laravel-doctrine/orm/blob/1.4/src/DoctrineServiceProvider.php#L251 should check if the auth manager has been bound before extending it.

### Actual behaviour

It doesn't check, so the auth manager is built from the container whether we need it or not

I've been trying to optimize an application and despite my best efforts this service provider is still the heaviest to register/boot. We don't use the Auth manager at all so the work done here is of no use to us, thus it would be great if it was optional, like `extendNotificationChannel`.

I can make a PR if this has a chance of getting accepted.

Jalle19:
Closes #302

- Only extend the auth manager if it has been bound

### DIFF
--- a/src/DoctrineServiceProvider.php
+++ b/src/DoctrineServiceProvider.php
@@ -248,21 +248,23 @@ class DoctrineServiceProvider extends ServiceProvider
      */
     protected function extendAuthManager()
     {
-        $this->app->make('auth')->provider('doctrine', function ($app, $config) {
-            $entity = $config['model'];
+        if ($this->app->bound('auth')) {
+            $this->app->make('auth')->provider('doctrine', function ($app, $config) {
+                $entity = $config['model'];
 
-            $em = $app['registry']->getManagerForClass($entity);
+                $em = $app['registry']->getManagerForClass($entity);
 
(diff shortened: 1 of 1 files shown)


Jalle19:
Will have to double-check whether this has the desired effect.

Jalle19:
Just verified with the debugger, seems to work as intended. The auth manager only gets extended if `$app->register(\Illuminate\Auth\AuthServiceProvider::class);` has been made during bootstrapping.

Jalle19:
Test failure is unrelated (fixed in https://github.com/laravel-doctrine/orm/pull/309)

Jalle19:
@patrickbrouwers any chance you could have a look at my PRs here and in the laravel-doctrine/migrations repository?

Jalle19:
Who should I ping to get this moving?

eigan:
You could try on Slack.

PR #309, opened 2018-02-14
[FIX] Fix intermittent unit test failures

Jalle19:
Sometimes the ID of the first "state" happens to be 3. Instead of
testing what it is not, specify an ID manually and test that the
ID equals that.


Jalle19:
Who should I ping to get this moving?

patrickbrouwers:
Nobody needs to be pinged. I pick stuff whenever I have time for it. Pinging me won't fasten that process.

Thanks for the PR though!

PR #311, opened 2018-02-17
[FEATURE]Update doctrine/orm to 2.6

sztyup:
- Use the recently relased 2.6 version of the doctrine ORM which has many improvements.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     ],
     "require": {
         "php": ">=7.0",
-        "doctrine/orm": "2.5.*",
+        "doctrine/orm": "2.5.*|2.6.*",
         "illuminate/auth": "5.5.*|5.6.*",
         "illuminate/console": "5.5.*|5.6.*",
         "illuminate/container": "5.5.*|5.6.*",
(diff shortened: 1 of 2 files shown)


patrickbrouwers:
@sztyup how well did you test this in your own app/a test app? Does everything still work, no changes on doctrine side that break anything on our side?

sztyup:
@patrickbrouwers I tested it on two of my own apps and as far as I can tell nothing breaks

sztyup:
Here are the possible BC breaks from the update, the only important is the php7.1 requirement, which should to be documented here too.
EDIT: I realized its better than that, for users with 7.0 it sticks with orm 2.5 and for 7.1 it updates to 2.6
https://github.com/doctrine/doctrine2/releases/tag/v2.6.0

xtrasmal:
Can we update?
I really like to use https://github.com/Ocramius/DoctrineBatchUtils which uses orm 2.6

patrickbrouwers:
Thanks for the PR, installing with 2.6 should be possible now. Will indeed install 2.5 if you are using <7.1

Issue #313, opened 2018-02-28
[BUG] Entity factory not persisting iterable objects correctly


garret-gunter:
Please prefix your issue with one of the following: [BUG] [PROPOSAL] [QUESTION].

### Package version, Laravel version
laravel-doctrine/orm 1.4.5, Laravel 5.5.35
### Expected behaviour
The entities returned from callables in a factory should persist if they are in Doctrine ArrayCollections or Laravel Collections.  Since the `entity()` function returns Laravel Collections when making multiple entities, the factory persisting logic should handle Collections.   

### Actual behaviour
The factory attempts to persist the ArrayCollection as if it was an entity itself.

### Steps to reproduce the behaviour
1. Make an entity with a one to many.   
2. Define a factory for the entity.  
3. Return a callable for the relationship.  
4. In the callable, return an instance of `\Doctrine\Common\Collections\ArrayCollection` or `\Illuminate\Support\Collection` from the callable.

I already have a fix for it.  I'll make a merge request for it.

garret-gunter:
PR merged.  Closing.

PR #314, opened 2018-02-28
[FIX] Check for iterable before object.

garret-gunter:

- Fix for #313

### DIFF
--- a/src/Testing/FactoryBuilder.php
+++ b/src/Testing/FactoryBuilder.php
@@ -213,18 +213,18 @@ class FactoryBuilder
         return array_map(function ($attribute) use ($attributes) {
             if ($attribute instanceof \Closure) {
                 $entity = $attribute($attributes);
-                if (is_object($entity)) {
+	            if (is_iterable($entity)) {
+		            foreach ($entity as $e) {
+			            if (is_object($e)) {
+				            $this->registry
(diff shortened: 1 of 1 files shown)


src/Testing/FactoryBuilder.php
```diff
@@ -213,18 +213,18 @@ protected function callClosureAttributes(array $attributes)
         return array_map(function ($attribute) use ($attributes) {
             if ($attribute instanceof \Closure) {
                 $entity = $attribute($attributes);
-                if (is_object($entity)) {
+	            if (is_iterable($entity)) {
```
patrickbrouwers: This is a php7.1 method, but the current branch also still allows 7.0

Perhaps we can do is_array || isntanceof Traversable , as that's basically what is_iterable does ?

patrickbrouwers (changes requested):

src/Testing/FactoryBuilder.php
garret-gunter: Fixed.  I always forget iterable was introduced in 7.1 and not 7.0.

patrickbrouwers:
Can you fix CS please.

garret-gunter:
Fixed cs.  Sorry for the delay.

patrickbrouwers (approved):

PR #339, opened 2018-07-21
[Fix] Logger to work with custom types

porozhnyy:
If using custom type(i using type https://github.com/Hunternnm/laravel-doctrine-odm/blob/master/src/JsonDocumentType.php from my mini-ODM) formatter crash if not exists `__toString` method, ignoring types method `convertToDatabaseValue`


porozhnyy:
@patrickbrouwers how do you look at such changes?

porozhnyy:
@patrickbrouwers could you see this PR?

PR #357, opened 2018-12-28
deprecate console commands that will be removed in Doctrine ORM 3

vv12131415:
Please prefix your pull request with one of the following: [FIX].

- deprecate `doctrine:convert:mapping`
- deprecate `doctrine:generate:entities`
- deprecate `doctrine:generate:repositories`

why
since [upgrade guide notices this](https://github.com/doctrine/doctrine2/blob/master/UPGRADE.md#bc-break-removed-code-generators-and-related-console-commands) and also [doctrine/DoctrineBundle](https://github.com/doctrine/DoctrineBundle/blob/master/Command/GenerateEntitiesDoctrineCommand.php#L79) does this

### DIFF
--- a/src/Console/ConvertMappingCommand.php
+++ b/src/Console/ConvertMappingCommand.php
@@ -41,6 +41,9 @@ class ConvertMappingCommand extends Command
      */
     public function handle(ManagerRegistry $registry)
     {
+        $this->error('The doctrine:convert:mapping command has been deprecated and will be removed in Doctrine ORM 3');
+
+
         $names = $this->option('em') ? [$this->option('em')] : $registry->getManagerNames();
 
         foreach ($names as $name) {
--- a/src/Console/GenerateEntitiesCommand.php
+++ b/src/Console/GenerateEntitiesCommand.php
@@ -39,6 +39,9 @@ class GenerateEntitiesCommand extends Command
      */
     public function handle(ManagerRegistry $registry)
     {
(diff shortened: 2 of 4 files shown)


vv12131415:
@eigan need your feedback

eigan:
Personally I never use these commands so I am okay. Dont understand why three of the commands actually need the deprecation notice though, its not like they are day-to-day commands to run. 

`MappingImportCommand.php` is possibly good for yaml users, since they actually need to do something (if they would like to upgrade, that is). Should be noted in documentation too (not sure if it is).

vv12131415:
`MappingImportCommand.php` is used for mapping from existing DB, so if users will create new mapping files with outdated driver (`yaml`) it will be strange, doesn't it?



> Should be noted in documentation too (not sure if it is).

I'm not sending PR to docs repo because it's noted about those changes in the base (`doctrine/orm`) repo

vv12131415:
@eigan if you think that I need to fix something, can you please write down, what should be fixed.
Or if everything is ok, than merge this PR please

eigan:
> MappingImportCommand.php is used for mapping from existing DB, so if users will create new mapping files with outdated driver (yaml) it will be strange, doesn't it?

Yes, agree. Keep it. 

> what should be fixed

https://github.styleci.io/analyses/XV5PdK?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #360, opened 2018-12-30
[FEATURE] update dependencies and tests

vv12131415:
- updated all the dependencies
- tests (since phpunit 5 differs so much from 7)
- use `laravel/framework` instead of `illuminate/*` since in the code you call `database_path`/`base_path`/etc. and that makes those functions undefined during development

Also, it should target 1.5 I think

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,8 +16,8 @@
         }
     ],
     "require": {
-        "php": ">=7.0",
-        "doctrine/orm": "2.5.*|2.6.*",
+        "php": "^7.2",
+        "doctrine/orm": "2.6.*",
         "illuminate/auth": "5.5.*|5.6.*|5.7.*",
         "illuminate/console": "5.5.*|5.6.*|5.7.*",
         "illuminate/container": "5.5.*|5.6.*|5.7.*",
@@ -27,16 +27,20 @@
         "illuminate/support": "5.5.*|5.6.*|5.7.*",
         "illuminate/validation": "5.5.*|5.6.*|5.7.*",
         "illuminate/view": "5.5.*|5.6.*|5.7.*",
-        "symfony/serializer": "^2.7|~3.0|~4.0"
+        "symfony/serializer": "^3.4|~4.0",
+        "itsgoingd/clockwork": "^3.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "~5.0",
+        "ext-json": "*",
         "mockery/mockery": "^1.0",
-        "barryvdh/laravel-debugbar": "~2.0",
-        "itsgoingd/clockwork": "~1.9",
         "illuminate/log": "5.5.*|5.6.*|5.7.*",
         "illuminate/notifications": "5.5.*|5.6.*|5.7.*",
-        "illuminate/queue": "5.5.*|5.6.*|5.7.*"
+        "illuminate/queue": "5.5.*|5.6.*|5.7.*",
+        "barryvdh/laravel-debugbar": "^3.2",
+        "phpunit/phpunit": "~7.0"
+    },
+    "config": {
+        "sort-packages": true
     },
     "autoload": {
         "psr-4": {
@@ -47,8 +51,11 @@
         ]
     },
     "autoload-dev": {
-        "classmap": [
-            "tests"
+        "psr-4": {
+            "LaravelDoctrine\\Tests\\": "tests/"
+        },
+        "files": [
+            "tests/functions.php"
         ]
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -18,15 +18,7 @@
     "require": {
         "php": "^7.2",
         "doctrine/orm": "2.6.*",
-        "illuminate/auth": "5.5.*|5.6.*|5.7.*",
-        "illuminate/console": "5.5.*|5.6.*|5.7.*",
-        "illuminate/container": "5.5.*|5.6.*|5.7.*",
-        "illuminate/contracts": "5.5.*|5.6.*|5.7.*",
-        "illuminate/pagination": "5.5.*|5.6.*|5.7.*",
-        "illuminate/routing": "5.5.*|5.6.*|5.7.*",
-        "illuminate/support": "5.5.*|5.6.*|5.7.*",
-        "illuminate/validation": "5.5.*|5.6.*|5.7.*",
-        "illuminate/view": "5.5.*|5.6.*|5.7.*",
+        "laravel/framework": "5.5.*|5.6.*|5.7.*",
         "symfony/serializer": "^3.4|~4.0",
         "itsgoingd/clockwork": "^3.1"
     },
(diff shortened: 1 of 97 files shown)


garret-gunter:
While php 7.0 has reached it's end of life, 7.1 is still supported.  So I don't think we should make the jump from supporting 7.0+ to just supporting 7.2.

composer.json
```diff
@@ -16,27 +16,23 @@
...
-        "illuminate/routing": "5.5.*|5.6.*|5.7.*",
-        "illuminate/support": "5.5.*|5.6.*|5.7.*",
-        "illuminate/validation": "5.5.*|5.6.*|5.7.*",
-        "illuminate/view": "5.5.*|5.6.*|5.7.*",
-        "symfony/serializer": "^2.7|~3.0|~4.0"
+        "php": "^7.2",
+        "doctrine/orm": "2.6.*",
+        "laravel/framework": "5.5.*|5.6.*|5.7.*",
+        "symfony/serializer": "^3.4|~4.0",
+        "itsgoingd/clockwork": "^3.1"
```
garret-gunter, line 23: Clockwork shouldn't be a dependency.

composer.json
```diff
@@ -16,27 +16,23 @@
...
+        "ext-json": "*",
         "mockery/mockery": "^1.0",
-        "barryvdh/laravel-debugbar": "~2.0",
-        "itsgoingd/clockwork": "~1.9",
         "illuminate/log": "5.5.*|5.6.*|5.7.*",
         "illuminate/notifications": "5.5.*|5.6.*|5.7.*",
-        "illuminate/queue": "5.5.*|5.6.*|5.7.*"
+        "illuminate/queue": "5.5.*|5.6.*|5.7.*",
+        "barryvdh/laravel-debugbar": "^3.2",
+        "phpunit/phpunit": "~7.0"
```
garret-gunter, line 32: `barryvdh/laravel-debugbar` should also not be a dependency.

composer.json
```diff
@@ -16,27 +16,23 @@
...
+    },
+    "config": {
+        "sort-packages": true
```
garret-gunter, line 35: Is sorting packages needed?

tests/DoctrineManagerTest.php
```diff
@@ -1,19 +1,21 @@
...
 
+namespace LaravelDoctrine\Tests;
+
 use Doctrine\Common\EventManager;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\DBAL\Connection;
 use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityManagerInterface;
 use Illuminate\Contracts\Container\Container;
+use InvalidArgumentException;
```
garret-gunter, line 11: I don't think importing from global namespace is needed.  If we really need to specify that it is from global namespace we can use the fully qualified name when using the name.  It is actually a small optimization to do it like that anyway.

e.g. `\InvalidArgumentException::class`

tests/EntityManagerFactoryTest.php
```diff
@@ -77,7 +82,7 @@ class EntityManagerFactoryTest extends PHPUnit_Framework_TestCase
     /**
      * @var array
      */
-    protected $caches = [ 'query', 'result', 'metadata' ];
+    protected static $caches = [ 'query', 'result', 'metadata' ];
```
garret-gunter, line 85: Is there a reason this is static now?  Static properties in tests can lead to weird or incorrect test results.

tests/Extensions/ExtensionManagerTest.php
```diff
@@ -141,15 +143,15 @@ public function test_boot_manager_with_one_manager_and_two_extensions()
         $this->container->shouldReceive('make')->with(ExtensionMock::class)->once()->andReturn(new ExtensionMock);
         $this->manager->register(ExtensionMock::class);
 
-        $this->container->shouldReceive('make')->with(ExtensionMock2::class)->once()->andReturn(new ExtensionMock2);
-        $this->manager->register(ExtensionMock2::class);
+        $this->container->shouldReceive('make')->with(\LaravelDoctrine\Tests\Mocks\ExtensionMock2::class)->once()->andReturn(new \LaravelDoctrine\Tests\Mocks\ExtensionMock2);
```
garret-gunter, line 146: It would be more consistent to import `\LaravelDoctrine\Tests\Mocks\ExtensionMock2` since `ExtensionMock` was imported.

tests/IlluminateRegistryTest.php
```diff
@@ -1,15 +1,19 @@
...
 use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\ORMException;
 use Illuminate\Contracts\Container\Container;
+use InvalidArgumentException;
 use LaravelDoctrine\ORM\EntityManagerFactory;
 use LaravelDoctrine\ORM\IlluminateRegistry;
 use Mockery as m;
 use Mockery\Mock;
+use stdClass;
```
garret-gunter, line 14: Same comment about global imports.  
> I don't think importing from global namespace is needed. If we really need to specify that it is from global namespace we can use the fully qualified name when using the name. It is actually a small optimization to do it like that anyway.
> 
> e.g. `\InvalidArgumentException::class`

tests/Loggers/Formatters/ReplaceQueryParamsTest.php
```diff
@@ -1,13 +1,17 @@
 <?php
 
+namespace LaravelDoctrine\Tests\Loggers\Formatters;
+
+use DateTime;
+use DateTimeImmutable;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
-use Doctrine\DBAL\Types\JsonArrayType;
 use Doctrine\DBAL\Types\Type;
+use Exception;
```
garret-gunter, line 9: Global import comment again.  

> I don't think importing from global namespace is needed. If we really need to specify that it is from global namespace we can use the fully qualified name when using the name. It is actually a small optimization to do it like that anyway.
> 
> e.g. `\InvalidArgumentException::class`

garret-gunter:
I made several of comments expressing my opinion concerning some of the changes.  The maintainers will have more comments and they may disagree with some of the things I brought up.  I only did a cursory glance since there are a lot of changes.    
  
  
I do think moving tests into namespaces is a good thing to do, but I'm not sure if the stub classes should be moved out of the same file.  Keeping them in the same file helps to keep the test self contained.

vv12131415:
@garret-gunter Thanks for comments, I will respond to them after the new year

eigan:
Changed target to `1.5` as lot of these changes was applied for the L6 PR which went into 1.5. 

There is several good changes here, so I would be great if you could rebase with 1.5.

vv12131415:
@eigan is there a way to make rebase by you? (fork from my repo, that's just idea how to do it).
As you can see I don't have much time to implement it.

eigan:
@vladyslavstartsev Yes. I can clone your fork and rebase. But I won't be able to update your fork I guess. You will need to do `git reset --hard eigan/update-dependencies`. _But since you don't have the time, it might be better to just start a new PR with the rebased commit._ 

I will let you know when I get the time, seems like a huge conflict.

vv12131415:
closing since too old

PR #380, opened 2019-06-03
[FIX] Readme updated to reflect support for Laravel 5.6, 5.7 and 5.8

jrgilman:
- Update the table in README.md in composer syntax to reflect support among the 1.4.x releases for various versions of Laravel 5.5 and greater.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Version | Supported Laravel Versions | Support
 1.1.x | 5.2.x |  EOL
 1.2.x | 5.2.x, 5.3.x | Security releases
 1.3.x | 5.4.x | Bugfix and security releases
-~1.4 | 5.5.x | New features
+~1.4.0 | 5.5.x | New features
 \>=1.4.3 | 5.6.x | New features
 \>=1.4.8 | 5.7.x | New features
 \>=1.4.10 | 5.8.x | New features
--- a/README.md
+++ b/README.md
@@ -50,10 +50,10 @@ Version | Supported Laravel Versions | Support
 1.1.x | 5.2.x |  EOL
 1.2.x | 5.2.x, 5.3.x | Security releases
 1.3.x | 5.4.x | Bugfix and security releases
-\>= 1.4.0 | 5.5.x | New features
-\>= 1.4.3 | 5.5.x, 5.6.x | New features
-\>= 1.4.8 | 5.5.x, 5.6.x, 5.7.x | New features
-\>= 1.4.10 | 5.5.x, 5.6.x, 5.7.x, 5.8.x | New features
+\>=1.4.0 | 5.5.x | New features
+\>=1.4.3 | 5.5.x, 5.6.x | New features
+\>=1.4.8 | 5.5.x, 5.6.x, 5.7.x | New features
+\>=1.4.10 | 5.5.x, 5.6.x, 5.7.x, 5.8.x | New features
 
 
 Require this package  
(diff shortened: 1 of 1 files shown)


eigan:
Not sure if we follow the same release schedule as Laravel, and only support the same versions.. If not, then you can probably disgard my comments. 

@vladyslavstartsev ?


_EDIT: I canceled my review because my assumption was wrong. Sadly the comments were then removed._

jrgilman:
@eigan just thought that it would be useful for people to know since the compatibility matrix at the moment makes it look like this package is severely outdated/no longer maintained, and would definitely cause people to skip over it. 

To build the compatibility matrix in this PR, I looked at the accepted commits named laravel 5.8, 5.7, and 5.6 to add the version of 1.4.x that is compatible with them.

README.md
```diff
@@ -50,7 +50,11 @@ Version | Supported Laravel Versions | Support
 1.1.x | 5.2.x |  EOL
 1.2.x | 5.2.x, 5.3.x | Security releases
 1.3.x | 5.4.x | Bugfix and security releases
-1.4.x | 5.5.x | New features
+\>=1.4.0 | 5.5.x | New features
+\>=1.4.3 | 5.5.x, 5.6.x | New features
+\>=1.4.8 | 5.5.x, 5.6.x, 5.7.x | New features
+\>=1.4.10 | 5.5.x, 5.6.x, 5.7.x, 5.8.x | New features
```
eigan: I don't think we need to repeat ourself in this column. Change to something like this: ?

Version | Supported Laravel Versions | Support
:---------|:----------|:----------
1.0.x |  5.1.x | EOL
1.1.x | 5.2.x |  EOL
1.2.x | 5.2.x, 5.3.x | Security releases
1.3.x | 5.4.x | Bugfix and security releases
\>=1.4.0 | 5.5.x | New features
\>=1.4.3 | 5.6.x | New features
\>=1.4.8 | 5.7.x | New features
\>=1.4.10 | 5.8.x | New features

eigan (changes requested):
This will be merged into 1.4. Could you do a PR for 1.5 too? :)

README.md
eigan: Actually, we should use `~` instead of `>=`, since 5.5 is not supported in 1.5.

https://getcomposer.org/doc/articles/versions.md#tilde-version-range-

jrgilman: Will do, I'll get this later today.

eigan: I have added a similar table here: https://github.com/laravel-doctrine/acl

jrgilman: Does 1.5 not support 5.6, 5.7 and 5.8 also, or just 5.5?

EDIT: If 1.5 does support 5.6 through 6.0 then the most recent push will work. I will make a PR into 1.5 once this one is accepted.

eigan: 1.5 is for ^6.0.

README.md
```diff
@@ -50,10 +50,10 @@ Version | Supported Laravel Versions | Support
 1.1.x | 5.2.x |  EOL
 1.2.x | 5.2.x, 5.3.x | Security releases
 1.3.x | 5.4.x | Bugfix and security releases
-\>=1.4.0 | 5.5.x | New features
-\>=1.4.3 | 5.5.x, 5.6.x | New features
-\>=1.4.8 | 5.5.x, 5.6.x, 5.7.x | New features
-\>=1.4.10 | 5.5.x, 5.6.x, 5.7.x, 5.8.x | New features
+~1.4 | 5.5.x | New features
```
eigan: Should be `~1.4.0`.

eigan:
Thank you @jrgilman :)

jrgilman:
@eigan :) I'm glad to help, thanks for putting up with my mistake! Will make a PR into 1.5 now.

PR #381, opened 2019-06-10
[FIX] Revert #289, Laravel 5.8 returns a boolean

eigan:
Previously Laravel would return void, but this changed in Laravel 5.8. So we can now use the proper value returned from store.


eigan:
Cannot merge this into 1.4 to fix 5.8, that would also break the other 5.x versions which uses 1.4.

PR #383, opened 2019-07-06
Remove unused utility ArrayUtil

### LINKED ISSUE #358: [QUESTION] use case of ArrayUtil
package 1.4
laravel 5.6

What is the use case of `ArrayUtil` class? since it's not used internally.
I mean it's easy to implement this method on my own

eigan:
Only had one method, which can be replaced by '??'

Closes #358

- Remove the ArrayUtil::get() utility

We should update the docs [here](https://github.com/laravel-doctrine/docs/blob/f3d829b098104975b896a2b7c60428d67e889850/orm/config-migrator.md), if this PR is merged.


eigan:
This is actually a breaking change, so should go into a `2.0` version.

Should probably have a `master` branch in this repo for next major release.

PR #384, opened 2019-07-06
Replace deprecated method calls

eigan:
# Changes proposed in this pull request:
- Replace usage of deprecated method.


eigan:
Not needed as it was fixed in L6 PR.

PR #386, opened 2019-07-06
[FIX] Remove custom json type which is a clone of the official

eigan:
This type was added at a time when the `json` type didnt exist in the Doctrine project.

The `json` type was later added to doctrine with the same fixes as we have here.
Overriding the official json type in Doctrine is not needed.

The JsonTypeTest proves this. Step through with Xdebug and you will notice that our custom json type will not be tested. `!Type::hasType('json')` is `false`, meaning our json type is not added.

**This should probably go into a 2.0 branch?**


garret-gunter:
This might be something that needs a deprecation notice before we remove it.  Some projects use that type since it's in the default configuration as an example.

If our json type has the exact behavior as the vanilla doctrine type, then it might be possible to just extend the base doctrine type and not override any of the methods.  Then we could deprecate the type and remove it in a later version.

eigan:
@garret-gunter agree. 

EDIT: Will fix this PR in the weekend. On vacation now.

config/doctrine.php
```diff
@@ -108,7 +108,6 @@
     |--------------------------------------------------------------------------
     */
     'custom_types'               => [
-        'json' => LaravelDoctrine\ORM\Types\Json::class
```
eigan, line 111: I removed this because this type is already registered by Doctrine, and our own is now deprecated.

https://github.com/doctrine/dbal/blob/2.7/lib/Doctrine/DBAL/Types/Type.php#L83

src/Types/Json.php
```diff
@@ -2,52 +2,17 @@
...
-     * Made to be compatible with Doctrine 2.4 and 2.5; 2.5 added getJsonTypeDeclarationSQL().
-     * {@inheritdoc}
-     */
-    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
-    {
-        if (method_exists($platform, 'getJsonTypeDeclarationSQL')) {
-            return $platform->getJsonTypeDeclarationSQL($fieldDeclaration);
-        }
-
-        return $platform->getClobTypeDeclarationSQL($fieldDeclaration);
```
eigan, line 30: This method was not implemented by `JsonArrayType`, which means it overrides `JsonType::getSQLDeclaration`. 


https://github.com/doctrine/dbal/blob/2.7/lib/Doctrine/DBAL/Types/JsonType.php#L42

They have the same behavior. Both will end up calling `AbstractPlatform::getJsonTypeDeclarationSQL`.

src/Types/Json.php
```diff
@@ -2,52 +2,17 @@
...
-     * When database value is null, we return null instead of empty array like our parent does.
-     * {@inheritdoc}
-     */
-    public function convertToPHPValue($value, AbstractPlatform $platform)
-    {
-        if ($value === null || $value === '') {
-            return null;
-        }
-
-        return parent::convertToPHPValue($value, $platform);
```
eigan, line 43: This method overrides the `JsonArrayType::convertToPHPValue()` and handles the `null` and `""` values. This is the main purpose of this type. 

Since we now remove this method and extend JsonType, lets compare the difference:

https://github.com/doctrine/dbal/blob/2.7/lib/Doctrine/DBAL/Types/JsonType.php#L68

The only difference is the check for invalid json value. Which means that when this PR is merged, json fields which have invalid json value will start throwing the `ConversionException` instead of `null`. I guess this should not happen unless you modify the column value yourself.

src/Types/Json.php
```diff
@@ -2,52 +2,17 @@
...
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getName()
-    {
-        return 'json';
```
eigan, line 51: `JsonType` also returns `json`. So no need to override this either.

eigan:
@garret-gunter I have readded our Json type without any of the methods. I also added some comments to help clarify why our type is no longer needed.

src/Types/Json.php
```diff
@@ -2,52 +2,17 @@
...
- * only arrays in mind. This extending type fixes a bug with the json_array type
- * wherein a null value in database gets converted to an empty array.
- * IMPORTANT NOTE: you must register custom types with Doctrine:
- *      \Doctrine\DBAL\Types\Type::addType('json', '\Path\To\Custom\Type\Json');
- * @link http://www.doctrine-project.org/jira/browse/DBAL-446
- * @link https://github.com/doctrine/dbal/pull/655
+ * @deprecated Extend JsonType directly
+ *
+ * This type was added when Doctrine only had `JsonArrayType`.
+ * `JsonArrayType` have a flaw with empty/null values.
```
garret-gunter: I think this should be 'had' instead of 'have'.
```suggestion
 * `JsonArrayType` had a flaw with empty/null values.
```

eigan: My english is not good, so you might be right, but JsonArrayType (`json_array`) does still have this issue _(though its deprecated)_.

tests/Types/JsonTypeTest.php
```diff
@@ -1,50 +0,0 @@
-<?php
-
-use Doctrine\DBAL\Platforms\AbstractPlatform;
-use Doctrine\DBAL\Types\Type;
-use LaravelDoctrine\ORM\Types\Json;
-use Mockery as m;
-use Mockery\Mock;
-
-class JsonTypeTest extends PHPUnit_Framework_TestCase
```
vladysta: I think, that currently you should leave this test, because we need to be sure that we haven't broke anything

eigan: As mentioned, this test have _no effect_. It just tests the JsonType from Doctrine.. I readded it anyway.

src/Types/Json.php
eigan: I have changed `have` to `had` now.

eigan:
So this PR just now started to fail on PHP 7.0, basically because of the inclusion of the test `JsonTypeTest` which I had previously deleted :-| 

The reason for it to fail is because we still support Laravel 5.5, which supports PHP 7.0.*. The JsonType which this PR propose extending, was added to doctrine/dbal in 2.6, which is only available for PHP 7.1.

So, this PR can only be merged when we no longer support Laravel 5.5. This version is an LTS and will stop receiving security fixes aug 2020.

vv12131415:
@eigan so right now, you can make PR which marks this type as deprecated?

eigan:
Yes, I can do that :+1:

eigan:
@vladyslavstartsev As I see it, we would need to release a new version (1.5) for Laravel 5.6? Marking as deprecated in 1.4 doesnt make sense because Laravel 5.5 users will not be able to replace our custom type with anything.

PR #363, opened 2019-01-21
Add factory callbacks

dallincoons:
This PR makes an update to include factory callbacks, introduced in 5.6 https://laravel.com/docs/5.7/database-testing#factory-callbacks

PR #389, opened 2019-08-05
[FIX] Revert changes to applying states in factory builder

garret-gunter:
- The merge request that added the new factory callbacks (#363) unintentionally broke factory states.  This merge request reverts the changes to applying states.

### DIFF
--- a/src/Testing/FactoryBuilder.php
+++ b/src/Testing/FactoryBuilder.php
@@ -272,7 +272,7 @@ class FactoryBuilder
     protected function applyStates(array $definition, array $attributes = [])
     {
         foreach ($this->activeStates as $state) {
-            if (! isset($this->states[$this->class][$state])) {
+            if (! isset($this->states[$state])) {
                 if ($this->stateHasAfterCallback($state)) {
                     continue;
                 }
@@ -298,7 +298,7 @@ class FactoryBuilder
      */
     protected function stateAttributes($state, array $attributes)
     {
-        $stateAttributes = $this->states[$this->class][$state];
+        $stateAttributes = $this->states[$state];
 
         if (! is_callable($stateAttributes)) {
             return $stateAttributes;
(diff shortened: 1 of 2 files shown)


eigan:
_I am not familiar with this part of laravel-doctrine. But I tried to look into the classes mentioned in this PR and the one mentioned._

Some notes:
- **Factory** can create `FactoryBuilder` instances, so it requires nested arrays with $className as key.
- **FactoryBuilder** builds entities for a single class.
- In **FactoryBuilder**, `$definitions`, `$afterMaking`, `$afterCreating` requires that you use nested arrays with class name as key. See the test: `FactoryBuilderTest::test_it_handles_after_making_callback` where it sends `$afterMaking`. It doesnt look like nesting is needed in this case.

This PR proposes removing the requirement of nesting with `$className` for `$states` in **FactoryBuilder**, but that would make FactoryBuilder a bit inconsistent. I think it would be better to change this behaviour in the entire FactoryBuilder. What do you think @garret-gunter?

Ping @dallincoons. Looks like you added the properties to FactoryBuilder. Do you have any input here?

iamrodrigo:
@eigan  I'm currently facing this problem with states on factory. 

As you mentioned, the ideal thing would be to remove the behaviour in the entire Factory or to remove [this line](https://github.com/laravel-doctrine/orm/blob/1.5/src/Testing/Factory.php#L281) and send all the states, simple fix and in fact that's how Eloquent does it https://github.com/illuminate/database/blob/master/Eloquent/Factory.php#L258

Let me know if you want me to create a PR for this, I need it in a project I'm currently working on so keen to do it asap :)

eigan:
I guess that would works too (how Eloquent does it), though a bit weird imo.

@noiarek @garret-gunter Can you take a look at my PR #410. I have implemented the changes suggested by @noiarek which will not cause any inconsistencies in Factory.

garret-gunter:
Those changes look good to me.  

I do agree that it is a bit weird.

garret-gunter:
Those changes look good to me.  

I do agree that it is a bit weird.

Issue #355, opened 2018-12-17
[PROPOSAL] Implement entity listner


vv12131415:
Implement something that described here https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/events.html#entity-listeners-class and here https://symfony.com/doc/current/bundles/DoctrineBundle/entity-listeners.html

garret-gunter:
You should be able to do this now.  Does adding it in annotations or other configurations not work?

vv12131415:
> Does it adding it in annotations or other configurations not

Haven't tried
I will. Thank you!

websmurf:
Just to add to this discussion, we are actively using listeners so I can confirm that this works already.

vv12131415:
@websmurf thank you!

mattkoch614:
Can someone post a code example?

PR #391, opened 2019-08-12
[FIX] BootChain: If the registry has already been booted then just run the callback

dpslwk:
Our Laravel application has got so big that something in the boot sequence is no longer happening in the expected order

The result of this is that `DoctrineServiceProvider` is being resolved before `AclServiceProvider` is booted and so our permissions are never setup

I have no idea if my fix in in the correct place but her it is.
BootChain keeps track of already resolved registries and if they are resolved it just runs the callback rather than storing it for later

- Immediately run onResolve callbacks if the ManagerRegistry has already been resolved

### DIFF
--- a/src/BootChain.php
+++ b/src/BootChain.php
@@ -12,11 +12,24 @@ class BootChain
     protected static $resolveCallbacks = [];
 
     /**
+     * @var ManagerRegistry[]
+     */
+    protected static $registries = [];
+
+    /**
      * @param callable $callback
      */
     public static function add(callable  $callback)
     {
-        static::$resolveCallbacks[] = $callback;
+        if (empty(static::$registries)) {
+            static::$resolveCallbacks[] = $callback;
+        } else {
+            foreach (static::$registries as $registry) {
+                if (is_callable($callback)) {
+                    call_user_func($callback, $registry);
+                }
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


eigan:
Sorry for the late response here @dpslwk. I need to be able to reproduce this bug before I can merge. 

Is `DoctrineServiceProvider` immediately followed by `AclServiceProvider` in your `providers` array in `config/app.php`?

dpslwk:
our project is here
https://github.com/NottingHack/hms2

issue shows up in my `#355-2fa` branch https://github.com/NottingHack/hms2/tree/%23355-2fa
which is working as composer is using my fork with this fix

`DoctrineServiceProvider` is not in our `config/app.php` as its picked up by Laravel's autoloader (via orm's composer file) but ACL is in there

Been a while since I looked at this but IIRC i did mess with `config/app.php` to no avail

`DoctrineServiceProvider` is being booted, registered, and resolved before `AclServiceProvider` is even booted

dpslwk:
looks like I'm failing travis 
I guess since I added `protected static $registries = [];` and that's not being cleared between tests
but Im not so familiar with testing to fix that in a hurry

eigan:
> DoctrineServiceProvider is being booted, registered, and resolved before AclServiceProvider is even booted

Have you tried setting a breakpoint in `BootChain::boot`? We need to find out what makes Laravel resolve the `ManagerRegistry` before Acl register() is called.

PR #402, opened 2019-10-16
[FIX] fixes #401

### LINKED ISSUE #401: [BUG] Double table prefixes for inherited associations
laravel-doctrine/orm: v1.5.0
doctrine/orm: 2.6.4
(tried to design minimal counter example without laravel)

### Expected behaviour
Using doctrine inheritance mapping (single or joined) with a Many-to-Many relation on a join table`join` together with a table prefix `p` should result in a table name `pjoin` in all queries.

### Actual behaviour
The table name in the queries in the above scenario is `ppjoin`, i.e. the prefix gets applied twice. The listener gets called twice, once for the base class and once for the child class.
This is especially bad since the command line tool orm:schema-tool:update uses the correct name `pjoin`, resulting in a runtime sql error when the queries use `ppjoin`

### Steps to reproduce the behaviour
As a minimal example use the following entities:

```
<?php
// Base.php
use Doctrine\ORM\Mapping;
/**
 * @Entity
 * @InheritanceType("SINGLE_TABLE")
 * @DiscriminatorColumn(name="discriminator", type="string")
 */
abstract class Base
{
  /** @Id @Column(type="integer") @GeneratedValue **/
  private $id;

  /**
   * @ManyToMany(targetEntity="Other")
   * @JoinTable(name="join")
   */
  private $others;
}
```
```
<?php
// Other.php
use Doctrine\ORM\Mapping;
/**
 * @Entity
 */
class Other
{
  /** @Id @Column(type="integer") @GeneratedValue **/
  private $id;
}
```
```
<?php
// Sub.php
use Doctrine\ORM\Mapping;
/**
 * @Entity
 */
class Child extends Base
{
}
```
Furthermore configure a table prefix `p`.
Then run `doctrine orm:schema-tool:create` resulting correctly in 3 tables (`pBase`, `pOther`, `pjoin`)
Then set/access `others` in a `Child` entity which results in a query that wants to use the non-existing table `ppjoin`.

blacktemplar:
Only add prefix for association tables if they are not inherited

### DIFF
--- a/src/Extensions/TablePrefix/TablePrefixListener.php
+++ b/src/Extensions/TablePrefix/TablePrefixListener.php
@@ -45,7 +45,8 @@ class TablePrefixListener implements EventSubscriber
         }
 
         foreach ($classMetadata->getAssociationMappings() as $fieldName => $mapping) {
-            if ($mapping['type'] == ClassMetadataInfo::MANY_TO_MANY && $mapping['isOwningSide']) {
+            if ($mapping['type'] == ClassMetadataInfo::MANY_TO_MANY && $mapping['isOwningSide']
+                    && (!array_key_exists("inherited", $mapping) || $mapping["inherited"] === $classMetadata->getName())) {
                 $mappedTableName                                                     = $mapping['joinTable']['name'];
                 $classMetadata->associationMappings[$fieldName]['joinTable']['name'] = $this->prefix . $mappedTableName;
             }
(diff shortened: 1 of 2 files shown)


eigan:
Would you be able to add a test for this? You might get some inspiration from `TablePrefixListenerTest::test_many_to_many_has_prefix`.

blacktemplar:
I added now a test case which fails without this pull request. Note, that the described situation strongly depends on the behaviour of the doctrine library and therefore the test simulates parts of a call to `Doctrine\ORM\Mapping\ClassMetadataFactory:addInheritedRelations` (see comment in code). I hope this is ok so for the unit test.

eigan:
Great! I will take a look tonight _(~7-8 hours)_ .

eigan:
Looks good to me @blacktemplar :)

Could you help me out and squash your commits? Seems like the Github UI might mess up the commit details. I want to make sure you get the credit for these changes. _Great if the commit messages would be updated to include "Fix #401: Double table prefixes for inherited associations"._

blacktemplar:
Ok, commits are squashed now.

eigan:
Thank you @blacktemplar :)

eigan:
Released as part of `1.4.11` and `1.5.1`.

PR #415, opened 2019-12-13
[FIX] use correct ManagerRegistry interface & pass tests

bounoable:
New version of doctrine/persistence broke the SubstituteBindings middleware and also the tests for the ServiceProvider. PR fixes the used ManagerRegistry interface.

PR #418, opened 2019-12-19
[FIX] fix doctrine/persistence upgrade

bounoable:
rebase of https://github.com/laravel-doctrine/orm/pull/415 for v1.4


eigan:
Looks like we instead must lock doctrine/persistence on 1.2 for the laravel-doctrine 1.4 version.. 

It's either that or stop supporting 7.0.

dpslwk:
larval dropped support for php 7.0 back in framework 5.6
and 6.x dropped support for php 7.1
doctrine/orm has required php 7.1 since 2.6
doctrine/persistence 1.2 or 1.3 is not tested against 7.0

eigan:
Thank you for testing this @bounoable. Unfortunately, since we still support PHP 7.0 in this release will we not be able to force doctrine/persistence 1.3. 

PHP 7.0 users installs doctrine/orm 2.5 which does not require doctrine/persistence. So we will never be able to fix this for laravel-doctrine/orm 1.4, unless we require at least php 7.1.

@dpslwk We support PHP 7.0 with laravel 5.5 and doctrine 2.5 in 1.4.

PR #429, opened 2020-02-03
[FIX] Change php7.4 snaphost to stable php7.4

KartaviK:


eigan:
I just merged 1.4 into 1.5, so this should not be necessary now because of #430 :)

PR #430, opened 2020-02-03
[FIX] Use stable php7.4 for 1.4 version

KartaviK:


eigan:
Thank you for this @KartaviK ! :)

PR #438, opened 2020-03-05
[FEATURE] Add support for Laravel 7.x

barryo:
All tests pass and `DoctrineTokenRepository` updated with tests.

Not tested on an application yet but changes are minor and should be
fine. Would be useful if someone else could test as I have no
production apps on Laravel 7 just yet.

No 1.6 or master branch to PR against.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 
 ```php
 $scientist = new Scientist(
-    'Albert', 
+    'Albert',
     'Einstein'
 );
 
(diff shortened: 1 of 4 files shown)


eigan:
Great work @barryo! :)

eigan:
The following packages also needs to release new versions for L7 support.

I will try to finish up the releases at the end of the week.

- [x] https://github.com/laravel-doctrine/migrations
- [x] https://github.com/laravel-doctrine/extensions
- [x] https://github.com/laravel-doctrine/acl

barryo:
Thanks for merging @eigan - I have open PRs for extensions and acl which should pass tests now that v1.6 is released. I missed migrations - if you don't get to it in the next couple days, I'll try.

eigan:
We need to upgrade to doctrine/migrations 2.0 for composer to be happy. Will take a look tomorrow.

eigan:
I did try to upgrade to migrations 2.0, but it went pretty bad. Lots of refactoring, so need to spend some time to get it right, which I currently do not have sadly.

PR #440, opened 2020-03-05
Added some flags to stop default extends.

rakshitbharat:
I have to use this package for Entity only to maintain blue print of database.

-
-
-

### DIFF
--- a/src/DoctrineServiceProvider.php
+++ b/src/DoctrineServiceProvider.php
@@ -49,8 +49,13 @@ class DoctrineServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        $this->extendAuthManager();
-        $this->extendNotificationChannel();
+        if($this->shouldExtendAuthManager()){
+            $this->extendAuthManager();   
+        }
+        
+        if($this->shouldExtendNotificationChannel()){
+            $this->extendNotificationChannel();
+        }
 
         if (!$this->isLumen()) {
             $this->publishes([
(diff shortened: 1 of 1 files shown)


eigan:
Could you provide some more information about what are you are trying to achieve with this change?

rakshitbharat:
I dont want auth or validator get's extended i want to use this package just for Mapping which useful for database management blueprinting

rakshitbharat:
in simple words i want to replace laravel migration with Mapping

eigan:
I still don't really understand, sorry. Are you using doctrine for _some_ entities, and Laravel for others? Like Users and notifications are handled by Laravel (Eloquent)? 

Which problems is adding the DoctrineAuthManager and DoctrineChannel causing for you?


Note however that it should be really simple for you to solve this:
1) Create your own ServiceProvider which extends DoctrineServiceProvider
2) Override the extendNotificationChannel() and extendAuthManager methods.
3) Use your custom ServiceProvider in config/app.php instead of DoctrineServiceProvider.

rakshitbharat:
really really nice idea.

You have also give me idea to override any other  package's provider.

your solution will work for me.

Thanks for this.

eigan:
Great! Rmember that there is no guarentee that these methods will continue to be named the way they do, or that the signature will not be changed to private or final. Our tests would not break by doing anything like this. I will try to remember that some folks like to override this class, but I will not stay arround here forever :)

I would recommend you add some tests that ensure that the DoctrineChannel and DoctrineAuthManager are not added. Not sure how, but I guess you will be able to figure that out. :)

PR #441, opened 2020-03-27
[FIX] Fixed callAfterCreating for one entity in FactoryBuilder::create

zergbz1988:

Fixed callAfterCreating for one entity in FactoryBuilder::create (laravel collect() method doesn't work correctly with one simple entity, but with array does)


eigan:
Thank you @zergbz1988! :)

eigan:
Sorry for the delay, released now with 1.5.8 and 1.6.1.

thicolares:
Thanks, @zergbz1988! I was debugging my factory and noticed that too. Then found your PR before starting a new one ;)

PR #443, opened 2020-05-05
[FEATURE] Implement custom keys and scoping in implicit bindings

ghostal:
Implement the features described in the Laravel manual for
the Doctrine route model bindings:

https://laravel.com/docs/7.x/routing#implicit-binding

I've tried my best to expand the existing functionality of the `LaravelDoctrine\ORM\Middleware\SubstituteBindings` middleware implementation to include the custom keys and scoping features described in the Laravel manual (link above), by adapting code in Laravel's own `Illuminate\Routing\ImplicitRouteBinding` class.

It's a bit hacky, as I was not at all familiar with Laravel's routing/binding system before looking at this, and not quite finished I think, but close. The code could do with some tidying up/refactoring, but more importantly currently there isn't support for the `UrlRoutable`/`getRouteKeyName` feature. I'm not sure where exactly that should slot into the logic.

Keen to hear any feedback/advice.

### DIFF
--- a/src/Middleware/SubstituteBindings.php
+++ b/src/Middleware/SubstituteBindings.php
@@ -6,7 +6,9 @@ use Closure;
 use Doctrine\ORM\EntityNotFoundException;
 use Doctrine\Persistence\ManagerRegistry;
 use Illuminate\Contracts\Routing\Registrar;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Routing\Route;
+use Illuminate\Support\Str;
 use LaravelDoctrine\ORM\Contracts\UrlRoutable;
 use ReflectionParameter;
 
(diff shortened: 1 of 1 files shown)


src/Middleware/SubstituteBindings.php
```diff
@@ -47,45 +49,91 @@ public function handle($request, Closure $next)
...
-                if ($parameter->getClass()->implementsInterface(UrlRoutable::class)) {
-                    $name = call_user_func([$class, 'getRouteKeyName']);
+            $parameterValue = $parameters[$parameterName];
 
-                    $entity = $repository->findOneBy([
-                        $name => $id
-                    ]);
-                } else {
-                    $entity = $repository->find($id);
-                }
```
ghostal, line 79: I'm wondering if this whole block should appear at line 104 of the new changes, instead of just `$entity = $repository->find($parameterValue);`. I think this would mean the `getRouteKeyName` logic only applies if a custom key hasn't been defined as part of the route parameters (e.g. `{user:username}`).

eigan:
Sorry for the lack of response here @ghostal. I will follow this PR and merge whenever you are happy with the result and some tests are added for the new functionality. Thanks for your contribution!

src/Middleware/SubstituteBindings.php
dpslwk: i think it should

Issue #451, opened 2020-07-14
[BUG][QUESTION] UrlRoutable contract conflicts with Illuminate\Contracts\Routing\UrlRoutable


dandrust-spark:
I'm trying to integrate Doctrine into my Laravel Project.  I'm running into an issue where my entities must implement two UrlRoutable contracts for routing to work properly.

I have a `Survey` entity whose route key needs to be it's `key` property rather than it's `id`.  In order for that to work I have o implement both UrlRoutable contracts:
* Implementing `LaravelDoctrine\ORM\Contracts\UrlRoutable` works in tandem with `LaravelDoctrine\ORM\Middleware\SubstituteBindings` to parse URLs for incoming requests.
* Implementing `Illuminate\Contracts\Routing\UrlRoutable` works in tandem with `Illuminate\Routing\RouteUrlGenerator` to generate URLs.

The issue is that both UrlRoutable contracts require implementation of `getRouteKeyName`.  The LaravelDoctrine version calls for this method to be static, where the Illuminate version expects this to be a member function.  A static and member function cannot be declared with the same name, and the `LaravelDoctrine\ORM\Middleware\SubstituteBindings` middleware explicitly checks to see if the interface is implemented on the Entity.  So a fast-and-loose duck-typing situation won't work.

I can work around this by passing the entity's route key to the route helper in the special circumstance but it would be nice if the contracts shared the same function as they're designed for the same purpose.

## Package version, Laravel version
laravel-doctrine/orm 1.6.0, laravel/framework v7.19.1

### Expected behaviour
I would expect the `getRouteKeyName` method in `LaravelDoctrine\ORM\Contracts\UrlRoutable` to match `Illuminate\Contracts\Routing\UrlRoutable`

### Actual behaviour
* `LaravelDoctrine\ORM\Contracts\UrlRoutable` declares `public static function getRouteKeyName(): string`
* `Illuminate\Contracts\Routing\UrlRoutable` declares `public function getRouteKeyName()`

### Steps to reproduce the behaviour
```php
<?php

namespace App\Entities;

use Illuminate\Contracts\Routing\UrlRoutable;
use LaravelDoctrine\ORM\Contracts\UrlRoutable as DoctrineUrlRoutable;

class Survey implements UrlRoutable, DoctrineUrlRoutable
{

    private $id;
    private $key;

    public static function getRouteKeyName() : string
    {
        return 'key';
    }

    public function getRouteKeyName()
    {
        return 'key';
    }

[...]

PR #454, opened 2020-07-28
[FEATURE] support Laravel UrlRoutable in SubstituteBindings

whizzrd:
- Support Laravel UrlRoutable in SubstituteBindings
- references #451

### DIFF
--- a/src/Middleware/SubstituteBindings.php
+++ b/src/Middleware/SubstituteBindings.php
@@ -7,7 +7,6 @@ use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\EntityNotFoundException;
 use Illuminate\Contracts\Routing\Registrar;
 use Illuminate\Routing\Route;
-use LaravelDoctrine\ORM\Contracts\UrlRoutable;
 use ReflectionParameter;
 
 class SubstituteBindings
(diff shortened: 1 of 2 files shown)


src/Middleware/SubstituteBindings.php
```diff
@@ -68,9 +67,15 @@ protected function substituteImplicitBindings(Route $route)
...
             if ($repository = $this->registry->getRepository($class)) {
-                if ($parameter->getClass()->implementsInterface(UrlRoutable::class)) {
+                if ($parameter->getClass()->implementsInterface(\LaravelDoctrine\ORM\Contracts\UrlRoutable::class)) {
                     $name = call_user_func([$class, 'getRouteKeyName']);
 
+                    $entity = $repository->findOneBy([
+                        $name => $id
+                    ]);
+                } elseif ($parameter->getClass()->implementsInterface(\Illuminate\Contracts\Routing\UrlRoutable::class)) {
+                    $name = call_user_func([new $class, 'getRouteKeyName']);
```
eigan: This will probably not work for everyone.

whizzrd:
Closing pull request due to discussion in #451

src/Middleware/SubstituteBindings.php
whizzrd: Yes, the same problem arises when implemeting `\Illuminate\Contracts\Routing\UrlRoutable` on an entity with a constructor requiring parameters, i'll have to research how Doctrine constructs entities without calling their constructor.

eigan: https://github.com/doctrine/orm/blob/2.7/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L913

Though I believe it would be better to experiment with the alternative pointed out here: https://github.com/laravel-doctrine/orm/issues/451#issuecomment-665449403 
Then implemeneting `Illuminate\Contracts\Routing\UrlRoutable` would not be needed.

whizzrd:
As stated by @eigan the better solution would be to [#451 (comment)](https://github.com/laravel-doctrine/orm/issues/451#issuecomment-665449403)

PR #462, opened 2020-10-15
[FIX] Change parameter order for php 8 compatibility

xaben:
This is a BC break as people may have extended the class.

If fixes the deprecation error produced by php 8:
`PHP Deprecated:  Required parameter $x follows optional parameter $y`

It fixes the issues inside the library, unfortunately this is not really relevant until doctrine team releases a php 8 compatible version, which might be 3.x, which in turn might require a lot more changes.

### DIFF
--- a/src/EntityManagerFactory.php
+++ b/src/EntityManagerFactory.php
@@ -119,13 +119,13 @@ class EntityManagerFactory
             $connection = (new MasterSlaveConnection($this->config, $connection))->resolve($driver);
         }
 
-        $this->setNamingStrategy($settings, $configuration);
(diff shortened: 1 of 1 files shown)


eigan:
doctrine have now released a PHP 8 supported version (2.8). :fireworks:

eigan:
Hi @xaben. After some discussion we landed on a different approach which should not break existing code. Released as a patch release to 1.5, 1.6 and 1.7. Please try it out if you can :) Thanks for you help on this! _(I think I will drop the 2.0 branch at some point soon)_

PR #471, opened 2020-12-22
[FIX] Update the Assert import

sstutz:
Fixes "Error: Class 'PHPUnit_Framework_Assert' not found"

PHPUnit introduced namespaces in v6.

- Update Assert import


eigan:
Thanks @sstutz  :+1:

PR #472, opened 2020-12-28
[FEATURE] support PrimaryReadReplicaConnection (former MasterSlaveConnection)

sserbin:
doctrine/dbal v2.11.0 (https://github.com/doctrine/dbal/pull/4054) has deprecated MasterSlaveConnection class in favour of PrimaryReadReplicaConnection
This change adds support for this new PrimaryReadReplicaConnection class
(while being BC-compatible with MasterSlaveConnection on older dbal versions)
and would allow users to use wrapperClass that extends PrimaryPrimaryReadReplicaConnection
(which otherwise fails as this package would try to use master/slave config instead of primary/replica one)


eigan:
Released with 1.7.6, thanks!

PR #473, opened 2021-01-15
[FIX] Failing PHP8 tests

Jurigag:

- Added reporting all errors in phpunit.xml (so all warnings during tests are reported, previously it was falsely passing tests on linux but failing on macos)
- Fixed bad optional parameter in `\LaravelDoctrine\ORM\DoctrineManager::extend()`
- Fixed bad optional parameter in `\LaravelDoctrine\ORM\DoctrineManager::callExtendOn()`
- Fixed bad optional parameter in `\LaravelDoctrine\ORM\DoctrineManager:::getMetaDataDriver()`
- Updated `\LaravelDoctrine\ORM\Middleware\SubstituteBindings::substituteImplicitBindings()` to use `ReflectionParameter::getType()` and `new \ReflectionClass()` intead of deprecated `\ReflectionParameter::getClass()`
- Updated dev dependency `itsgoingd/clockwork` to `^5.0` to eliminate `Required parameter $message follows optional parameter $level` warning
- Fixed bad optional parameters in \LaravelDoctrine\ORM\EntityManagerFactory`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,10 +36,10 @@
         "phpunit/phpunit": "^9.3",
         "mockery/mockery": "^1.3.1",
         "barryvdh/laravel-debugbar": "~3.0",
-        "itsgoingd/clockwork": "~1.9",
         "illuminate/log": "^8.0",
         "illuminate/notifications": "^8.0",
-        "illuminate/queue": "^8.0"
+        "illuminate/queue": "^8.0",
+        "itsgoingd/clockwork": "^5.0"
     },
     "autoload": {
         "psr-4": {
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -14,6 +14,11 @@
             <directory suffix="Test.php">./tests/</directory>
         </testsuite>
     </testsuites>
+    <php>
+        <ini name="display_errors" value="On" />
+        <ini name="display_startup_errors" value="On" />
+        <ini name="error_reporting" value="E_ALL" />
+    </php>
     <filter>
         <whitelist>
             <directory suffix=".php">src/</directory>
--- a/src/DoctrineManager.php
+++ b/src/DoctrineManager.php
@@ -44,7 +44,7 @@ class DoctrineManager
      * @param string|null     $connection
      * @param string|callable $callback
      */
-    public function extend($connection = null, $callback)
+    public function extend($connection, $callback)
     {
         $this->onResolve(function (ManagerRegistry $registry) use ($connection, $callback) {
             $this->callExtendOn($connection, $callback, $registry);
@@ -68,7 +68,7 @@ class DoctrineManager
      * @param string|callable $callback
      * @param ManagerRegistry $registry
      */
-    private function callExtendOn($connection = null, $callback, ManagerRegistry $registry)
+    private function callExtendOn($connection, $callback, ManagerRegistry $registry)
     {
         $manager = $registry->getManager($connection);
 
@@ -130,7 +130,7 @@ class DoctrineManager
      * @param  ManagerRegistry $registry
      * @return MappingDriver
      */
-    public function getMetaDataDriver($connection = null, ManagerRegistry $registry)
+    public function getMetaDataDriver($connection, ManagerRegistry $registry)
     {
         $registry = $registry ?: $this->container->make('registry');
         $manager  = $registry->getManager($connection);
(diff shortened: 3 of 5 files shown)


Jurigag:
Also, can I do the same for 1.5 version and 1.6 version to add support of php8? Though it will be kind of problematic to keep backward compatibility due to those parameter order.

We are actually staying on laravel 6 on our project, that's why i ask.

dpslwk:
Breaking changes should not be introduced in minor versions

If that means a upper php version limit so be it (my view)

Laravel 6.x is LTS till September 2022
PHP 7.4 has security till November 2022
So not being able to have php 8 with L6.x is not a support issue

Jurigag:
> Breaking changes should not be introduced in minor versions
> 
> If that means a upper php version limit so be it (my view)
> 
> Laravel 6.x is LTS till September 2022
> PHP 7.4 has security till November 2022
> So not being able to have php 8 with L6.x is not a support issue

But you know that Laravel 6.x support PHP 8 as well? This is why i am not sure why support of it was dropped. I mean currently laravel-doctrine/orm is stopping us from update to php8 on laravel 6.

Jurigag:
Also changing here https://github.com/laravel-doctrine/orm/blob/1.5/src/DoctrineManager.php#L71 this optional parameter as first one to be a required one won't be a breaking change, because it was actually a bug anyway. Since this version is supported only from php 7.2 it wouldn't be possible anyhow to not pass this first argument anyway. Check example code here, it's fatal error from PHP 7.1:

https://3v4l.org/YZEd7

eigan:
I don't have much time pr day to work on this, but I just want to let you guys know that Iam thinking about it and watching every PR.  As you have pointed out @Jurigag, we don't really need to make this a breaking change.

dpslwk:
> But you know that Laravel 6.x support PHP 8 as well? This is why i am not sure why support of it was dropped. I mean currently laravel-doctrine/orm is stopping us from update to php8 on laravel 6.

Because Laravel 6.x was able to add php 8 support with out a BC

> Also changing here https://github.com/laravel-doctrine/orm/blob/1.5/src/DoctrineManager.php#L71 this optional parameter as first one to be a required one won't be a breaking change, because it was actually a bug anyway. Since this version is supported only from php 7.2 it wouldn't be possible anyhow to not pass this first argument anyway. Check example code here, it's fatal error from PHP 7.1:
> 
> https://3v4l.org/YZEd7

For me the bug fix to that would sort of issue should be to remove the `= null` not change the param order, that to me would not be a BC
(which is the fix that should have likely been applied at the point when support for php < 7.1 was dropped)

Jurigag:
Okay cool, I can then keep order of parameters here as it was, should I update this to previous state?

dpslwk:
> Okay cool, I can then keep order of parameters here as it was, should I update this to previous state?

not sure as these BC's also need to be considered
https://github.com/laravel-doctrine/orm/pull/462


Should a correct approach to all this be have our 2.x branch fix the order of all bad methods to correctly allow optional params and aim to release it with support for Laravel 9.x in ~March (adding a quick UPGRADE.md to list all the bc method changes)
but have the 1.5>1.7 branches remove the false optional to allow php8 support

eigan:
I think we should just try out these changes in 1.5 (including #462). 

The order of the parameters should stay as is. The type should be removed. This will allow null and not introduce any breaking change. When done, can we merge it into 1.5, 1.6, and 1.7 for L6-L8 support.

Should also look into why laravel-doctrine/orm 1.6/1.7 was required for laravel-doctrine/migrations 2.0. Not allowing 1.5 will make it impossible to support PHP 8 on L6.

These are the stuff remaining.

Jurigag:
@eigan @dpslwk 

I reverted changing parameter order, just removed it as optional and it should be enough for this because this is not breaking change(it was just a bug). I will try to make same changes to 1.5,1.6,1.7 so PHP8 is supported

eigan:
@Jurigag Note that I merge 1.5 into 1.6, 1.7 and so on. So if you fix it in 1.5 first, then it might be done for the rest of the versions too :)

Jurigag:
Oh cool, will do that. Only for 1.5 then.

Jurigag:
Created PR for 1.5, but travis seems to have issue.

PR #475, opened 2021-01-22
Add PHP8 support

Jurigag:

- Add PHP8 support


Jurigag:
@eigan Not sure why but travis doesn't start

eigan:
Changed to github actions. Could you rebase and try again?

Jurigag:
@eigan done :)

eigan:
Great work! 

Why was the `FactoryTest` removed?

Jurigag:
It was using assertAttributeEquals from older phpunits to check private/protected members, where tests shouldn't actually test implementation details, only public methods.

But as i see in branch 1.7/2.0 you rewritten this test to use other method, will do the same.

Jurigag:
@eigan done

eigan:
Thanks! 

I will do some more testing before releasing this. Will also need to solve laravel-doctrine/migrations, think this might need to be solved in 2.0 (allow only L6, let 2.1 deal with L7).  Will experiment this weekend. Let me know if you are trying to solve it. 
(I am also available on slack).

Jurigag:
Cool, http://slack.laraveldoctrine.org returns 403 @eigan I will for sure check it on 1.5.x-dev if tests are passing for us and let here update.

Jurigag:
When to expect new release? @eigan

eigan:
Did you get it through your tests? :)

Currently preparing laravel-doctrine/migrations
https://github.com/laravel-doctrine/migrations/pull/107

I will try to get someone to look into the slack issue. Our website is not something I have control over.

Jurigag:
We have a few more packages needed to update to work with PHP8, will let you know.

PR #480, opened 2021-02-15
[FIX]? Suggest limiting vendor:publish command to current package in docs.

JamesAnelay:

Changes the command in the readme so it doesn't publish other dependancies config files if they are not needed.


eigan:
Thanks :+1:

PR #485, opened 2021-02-28
[FEATURE] Added setQuoteStategy method to EntityManagerFactory

smarcet:
added method to EntityManagerFactory to allow to set the used
quote strategy (defaults to DefaultQuoteStrategy).
set by configuration key "quote_strategy".

Signed-off-by: smarcet <smarcet@gmail.com>

- Allows to set quote_strategy as state by doctrine doc https://www.doctrine-project.org/projects/doctrine-orm-module/en/3.1.x/configuration.html#how-to-use-quote-strategy

### DIFF
--- a/src/EntityManagerFactory.php
+++ b/src/EntityManagerFactory.php
@@ -120,6 +120,7 @@ class EntityManagerFactory
         }
 
         $this->setNamingStrategy($settings, $configuration);
+        $this->setQuoteStrategy($settings, $configuration);
         $this->setCustomFunctions($configuration);
         $this->setCustomHydrationModes($configuration);
         $this->setCacheSettings($configuration);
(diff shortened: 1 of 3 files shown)


src/EntityManagerFactory.php
```diff
@@ -318,6 +320,19 @@ protected function setNamingStrategy(array $settings, Configuration $configurati
...
 
+    /**
+     * @param array         $settings
+     * @param Configuration $configuration
+     */
+    protected function setQuoteStrategy(array $settings, Configuration $configuration)
+    {
+        $strategy = Arr::get($settings, 'quote_strategy', DefaultQuoteStrategy::class);
+
+        $configuration->setQuoteStrategy(
```
eigan, line 332: Could you instead not set quoting strategy if $strategy === null? This will keep the existing behavior and if Doctrine decides to deprecate DefaultQuoteStrategy for some reason then we do not have to do anything.

PR #490, opened 2021-05-06
[FIX] Missing check for builtIn type in SubstituteBindings middleware

egorbwork:
For typed parameters in controllers SubstituteBindings is failing with message `Class 'type' does not exist `(where type can be string, int, etc.). To fix it was required the check of the ReflectionNamedType isBuiltin method. Added required changes and added unit test for this purpose.

### DIFF
--- a/src/Middleware/SubstituteBindings.php
+++ b/src/Middleware/SubstituteBindings.php
@@ -109,7 +109,7 @@ class SubstituteBindings
     {
         $class = null;
 
-        if (($type = $parameter->getType()) && $type instanceof \ReflectionNamedType) {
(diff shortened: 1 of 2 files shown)


egorbwork:
This changes should be available for all php 8 versions of the package, so I changed destination to 1.5 branch. If I should change, it please inform me.

eigan:
Thanks @egorbwork! I will merge and release this tomorrow morning (in about 7-8 hours).

eigan:
Released with 1.5.11, 1.6.5, and 1.7.7 🙂

egorbwork:
Thanks a lot!

PR #491, opened 2021-05-18
[FEATURE] Add pointers to non-default connection settings

laurentvd:
- Following up on https://github.com/laravel-doctrine/migrations/issues/112, I propose to add a hint in the config docblock, pointing users in the right direction. I tried to keep the wording concise (and of course follow Taylor's layout of 3 chars less per line, for good measure), but feel free to make suggestions.

If desired, I can add similar documentation to the laravel-doctrine/docs repo.

### DIFF
--- a/config/doctrine.php
+++ b/config/doctrine.php
@@ -17,6 +17,10 @@ return [
     | Available connections: mysql|oracle|pgsql|sqlite|sqlsrv
     | (Connections can be configured in the database config)
     |
+    | Depending on chosen the database connection, various other settings are
+    | available. Check the available settings for your connection type in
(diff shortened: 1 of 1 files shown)


eigan:
Looks good. Thanks 👍🏻 

_Sorry for the delay._

laurentvd:
@eigan I just re-read my commit and noticed the order of two words is wrong, despite reading a couple times over when I made the PR. I'm sorry for this. 

"Depending on chosen the database connection" should've been "Depending on the chosen database connection".

Shall I create a new PR or do you want to make the change?

Again, my apologies for this :/

eigan:
I should have noticed that too :) I can fix it directly, no worries :)

Issue #494, opened 2021-06-30
[PROPOSAL] Add attributes metadata support


dretsa:
Please prefix your issue with one of the following: [BUG] [PROPOSAL] [QUESTION].

### Package version, Laravel version
doctrine/orm:2.9.1
laravel-doctrine/orm: 1.7.7

### Expected behaviour
Users must be able to specify the new attributes driver when using PHP 8+

### Actual behaviour
There is no metadata driver in the package.

### Steps to reproduce the behaviour
1. Create new Laravel project
2. Install laravel-doctrine/orm
3. Try to use #[ORM\Entity] and #[ORM\Table]
4. You get an error `Class "....." is not a valid entity or mapped super class.`

eigan:
Your PR have been released with 1.5.13, 1.6.7 and 1.7.9. Please try it out. Thanks 👍🏻

jurgenbosch:
@eigan This does not seem to work with the migrations:diff command. Are you aware of that?

PR #495, opened 2021-06-30
[FEATURE] Add attributes metadata support [Issue #494]

dretsa:
- Add new attributes metadata support

### DIFF
--- a/src/Configuration/MetaData/Attributes.php
+++ b/src/Configuration/MetaData/Attributes.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace LaravelDoctrine\ORM\Configuration\MetaData;
+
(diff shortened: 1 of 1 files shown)


dretsa:
The only thing that needs to be checked is that this commit doesn't throw errors because `AttributeDriver` was added in `doctrine/orm:2.9`, but composer.json specifies `^2.6|^2.7|^2.8`, which AFAIK would behave the same as `^2.6`.

eigan:
Thanks @colorninja 🙂 

I have vacation but will take a look at this when I get the time.

eigan:
@colorninja Could you rebase on 1.5 and set target to 1.5? :)

eigan:
Nevermind. I just cherry-picked you commit instead. Much easier. Thanks!

PR #499, opened 2021-09-06
Fixed handling of database connections configuration for "read/write" replicas

### LINKED ISSUE #498: [BUG] Replicas configuration is not handled correctly in Laravel 8
Package 1.7.9
Laravel 8.0

### Expected behaviour
Laravel 5.6 format for replicas config should be handled by package.
Here is the format from [the doc](https://laravel.com/docs/8.x/database#read-and-write-connections):
```php
'mysql' => [
    'read' => [
        'host' => [
            '192.168.1.1',
            '196.168.1.2',
        ],
    ],
// ...
```
### Actual behaviour
Now it fails for two reasons:
1. Because it expect config [in laravel 5.5 format](https://laravel.com/docs/5.5/database#read-and-write-connections), where host is a plain value.
```php
'mysql' => [
    'read' => [
        'host' => '192.168.1.1',
    ],
  // ...
],

2. Because it expects 'read' key to be an array of configs like this:
```php
'read' => [
                [
                    'port'     => 3308,
                    'database' => 'test2',
                    'host'     => 'localhost',
                    'username' => 'homestead',
                    'password' => 'secret'
                ],
                [
                    'host'     => 'localhost2',
                    'port'     => 3309,
                    'database' => 'test',
                    'username' => 'homestead',
                    'password' => 'secret'
                ],
            ],
```
which is actually a not documented thing in laravel.
So documented way is not supported, and undocumented is. Which is weird

bravik:
Fixes https://github.com/laravel-doctrine/orm/issues/498

### DIFF
--- a/src/Configuration/Connections/PrimaryReadReplicaConnection.php
+++ b/src/Configuration/Connections/PrimaryReadReplicaConnection.php
@@ -40,11 +40,19 @@ class PrimaryReadReplicaConnection extends Connection
     {
         $driver = $this->resolvedBaseSettings['driver'];
 
+        $writeReplicas = $this->getReplicasConfiguration(isset($settings['write']) ? $settings['write'] : [], $driver);
+
+        if (count($writeReplicas) !== 1) {
+            throw new \InvalidArgumentException(
+                "There should be exactly 1 write replica. " . count($writeReplicas) . " found."
+            );
+        }
+
         $resolvedSettings = [
             'wrapperClass'  => $settings['wrapperClass'] ?? PrimaryReadReplicaDoctrineWrapper::class,
             'driver'        => $driver,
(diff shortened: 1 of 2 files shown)


bravik:
@sserbin can you have a look as an author of affected code, please?

sserbin:
> @sserbin can you have a look as an author of affected code, please?

I think this should work, in principal :+1: I don't have approval permissions here though so I'll leave that to maintainers

eigan:
@bravik thanks! Looks good. Could you squash your commits and write a better messages containing what your are doing?

bravik:
> @bravik thanks! Looks good. Could you squash your commits and write a better messages containing what your are doing?

Done. Looks better now?

eigan:
Thanks. Will release this soon 👍🏻

eigan:
Released with 1.7.10. Thanks!

PR #501, opened 2021-09-20
[FEATURE] Add Laravel event QueryExecuted support (Sentry support)

FandaOG:
- Laravel standard event support - [QueryExecuted](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Events/QueryExecuted.php)
- supports [Sentry](https://github.com/getsentry/sentry-laravel/blob/c9077d6765bff3f7f8ed7e91774d5bd224847a2b/src/Sentry/Laravel/Tracing/EventHandler.php#L217) (Sentry bind QueryExecuted event)


src/Loggers/SQL/DoctrineSQLLogger.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function stopQuery()
+    {
+        $executionTime = microtime(true) - $this->start;
+        // TODO get connection name
+        // this class is Laravel connection wrapper (QueryExecuted needs getName())
+        $object                    = new class {
```
eigan: This should actually be a `\Illuminate\Database\Connection` instance. Though implementing that seems like a pretty huge task.

eigan: The laravel-telescope QueryWatcher uses the ->connection->prepareBindings on this event: https://github.com/laravel/telescope/blob/e39423d3ee18898eeb1e0adf2fbe7b74a4cc4996/src/Watchers/QueryWatcher.php

FandaOG: We have 2 options i guess. 

1. extend my anonymous object (add prepareBindings)
2. try to implement [ConnectionInterface](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/ConnectionInterface.php) with empty methods. This methots could be extended in future if it will be nessesery.

I will try the 2nd option. Create a wrapper for DoctrineConnection with Laravel ConnectionInterface.

FandaOG: @eigan I've maybe found a way. Please check my last commit https://github.com/laravel-doctrine/orm/pull/501/commits/02080735013911f9185964dbab3911d34d8a62d3

eigan: Cool. Could you make some tests for this too?

FandaOG: > Cool. Could you make some tests for this too?

I've check Sentry support and it works.

eigan: I was thinking about PHPUnit test.

rosamarsky:
I would do it more cleanly without public properties and with DI for EventDispatcher mocking ability:
https://gist.github.com/rosamarsky/18a6ed0786179c39e8eb3cc6d3ae953a
There is also need to make full filled **Illuminate Connection** from **Dbal Connection** before pass it into event.

But this feature is not a Logger. You are dispatching an event, which means that **Illuminate Query** has been executed and some applications and packages handle this event, and in some cases this can be fatal. So, IMHO, this feature makes sense, but this is not a logger. You can use the Sentry SDK to create a some **SentryLogger** and send queries through it...

src/Loggers/SQL/DoctrineSQLLogger.php
FandaOG: I've added few tests.

FandaOG:
@rosamarsky Thank you for your proposal. I used them.

I know it is not a Logger. But I think it fits there. It is optional developers can enable it optionaly. Implement this like a standard event is better because it could support more libraries, solutions... So it solve Sentry SDK support (error tracing and performance monitoring) and it could support Laravel Telescope etc.

src/Loggers/SQL/DoctrineSQLLogger.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function startQuery($sql, ?array $params = null, ?array $types = null)
+    {
+        $this->start  = microtime(true);
+        $this->query  = $sql;
+        $this->params = $params;
```
superbiche, line 46: When using an array as binding (for IN clauses) this will break: 
> PDO::quote(): Argument #1 ($string) must be of type string, array given 

Telescope `QueryWatcher` tries to quote values that are not int/float: 
```
if ($binding === null) {
  $binding = 'null';
} elseif (! is_int($binding) && ! is_float($binding)) {
  $binding = $this->quoteStringBinding($event, $binding);
}
```

Using a `join()` makes it work while keeping the query readable: 
image

TomHAnderson:
Closing due to lack of followup.

PR #502, opened 2021-09-22
[FIX] Pass the Throttle config value to DoctrineTokenRepository

yusrub:
[FIX]
- In Laravel 8, we can control the throttle value for forgot passwords in the file config/auth.php 
```
'passwords' => [
        'users' => [
            'throttle' => 30,
        ],
    ],
```
- When register the DoctrineTokenRepository using PasswordBrokerManager, currently, the throttle value is ignored.
- I applied the fix to take the throttle value if its set in config/auth.php

### DIFF
--- a/src/Auth/Passwords/PasswordBrokerManager.php
+++ b/src/Auth/Passwords/PasswordBrokerManager.php
@@ -19,7 +19,8 @@ class PasswordBrokerManager extends \Illuminate\Auth\Passwords\PasswordBrokerMan
             $this->app->make('registry')->getConnection($connection),
             $config['table'],
             $this->app['config']['app.key'],
-            $config['expire']
+            $config['expire'],
+            $config['throttle'] ?? 0
         );
     }
 }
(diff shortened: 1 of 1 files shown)


src/Auth/Passwords/PasswordBrokerManager.php
```diff
@@ -19,7 +19,8 @@ protected function createTokenRepository(array $config)
             $this->app->make('registry')->getConnection($connection),
             $config['table'],
             $this->app['config']['app.key'],
-            $config['expire']
+            $config['expire'],
+            $config['throttle'] ?? 0
```
eigan: It is 60 for all users today (default when not supplied). But this patch would set it to zero. Should we use `?? 60` instead perhaps?

yusrub: Yes, we can change it to default 60.

eigan:
Thanks!

eigan:
Released with 1.7.11 now.

Issue #506, opened 2021-12-10
[BUG] The `Auth\Passwords\DoctrineTokenRepository` creates tokens without hashing


rosamarsky:
I used the default Illuminate Password Broker before I found in your documentation [This page](http://laraveldoctrine.org/docs/1.4/orm/passwords)
After changing `PasswordResetServiceProvider` in config/app.php I checked the table `password_resets` and saw not hashed tokens. It looks really unsafe. 

<img width="801" alt="image" src="https://user-images.githubusercontent.com/15215861/145599255-320c7cab-22b0-4ef5-9d55-24bf2911e01c.png">

_Package version 1.4, Laravel version 8.0_

dpslwk:
I believe that that token is generated here 
https://github.com/laravel-doctrine/orm/blob/2824a8516ac8b5b812d7195f01c7c64470abc179/src/Auth/Passwords/DoctrineTokenRepository.php#L213-L221

which just results on a different Hash/Token that laravel which is likely using there built in Hash Facade

dpslwk:
ah laravel are building it the same but are pushing the token through the hasher before it goes into the database

https://github.com/laravel/framework/blob/cf00a5621f2eacdb978d3176e6d601193d9dc57d/src/Illuminate/Auth/Passwords/DatabaseTokenRepository.php#L207-L215

https://github.com/laravel/framework/blob/cf00a5621f2eacdb978d3176e6d601193d9dc57d/src/Illuminate/Auth/Passwords/DatabaseTokenRepository.php#L111-L121


Happy to take  PR to fix this, but it might be considered a BC, since it would invalidate all existing tokens

PR #508, opened 2021-12-13
[FIX] Hashing password-reset tokens before storing

rosamarsky:
[FIX] Fix for issue #506. Hashing password-reset tokens before storing.

### DIFF
--- a/src/Auth/Passwords/DoctrineTokenRepository.php
+++ b/src/Auth/Passwords/DoctrineTokenRepository.php
@@ -7,6 +7,7 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Schema\Table;
 use Illuminate\Auth\Passwords\TokenRepositoryInterface;
 use Illuminate\Contracts\Auth\CanResetPassword;
+use Illuminate\Contracts\Hashing\Hasher as HasherContract;
 use Illuminate\Support\Str;
 
 class DoctrineTokenRepository implements TokenRepositoryInterface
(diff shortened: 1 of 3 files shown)


eigan:
As @dpslwk pointed out, this is invalidating all existing tokens. Though we never release a new major so cannot really see how else we can solve this...

rosamarsky:
@eigan, maybe you can create a dev branch with all backward incompatible commits and tag it with git like `dev-master` and merge it someday with a new major version. 

BTW, I don't see a problem if this commit invalidates all tokens. You can just use the password recovery request again to get a new link with a new token.

dpslwk:
I believe `src/Auth/Passwords/PasswordBrokerManager.php` also needs an update to pass the new constructor param

dpslwk:
oh and looks like we should have also done this change 
https://github.com/laravel/framework/pull/13270/files

back around  v5.2.31 (2016-04-27), as the change to APP_KEY gen was introduced in v5.2.26 (2016-03-25)

rosamarsky:
@dpslwk sorry, I forgot to add this file to the commit

dpslwk:
> @dpslwk sorry, I forgot to add this file to the commit

no worries its why we have code review 😉

eigan:
Thank you @rosamarsky and @dpslwk!


Going for a release announcement instead of bc. Sorry in advance for anyone getting angry users over this.

dpslwk:
make some sense, I mean these a reset tokens they should only have a short lifetime anyway so effects should only be during a short transition period

PR #510, opened 2021-12-18
[FIX] Trying to print_r circular entity dependency results in OOM in InteractsWithEntities 

k0ka:
### Problem
The `InteractsWithEntities` trait has `entitiesMatch`/`noEntitiesMatch` assertions which accept criteria as an argument. This criteria might contain other entities: for example if we are searching by a `ManyToOne` relation. The code outputs criteria in `print_r`. But in a complex setup where `entityA` has the relation to `entityB` and `entityB` has the relation to `entityA` this may end up with a circular dependency and fail with the PHP error `Allowed memory size of X bytes exhausted `

- All entities in `criteria` in `InteractsWithEntities` are replaced by their ids. 
- simple tests for `InteractsWithEntities`


eigan:
Thanks! :)

PR #513, opened 2022-01-18
[FIX] EntityGenerator - Strict types error fix

DocentBF:
Command doctrine:generate:entities causes TypeError. EntityGenerator::setNumSpaces has int attribute with strict_types enabled.

- Call EntityGenerator::setNumSpaces with attribute type conversion


eigan:
Thanks! I cherry-picked your commit into 1.5, and will merge it up into 1.7 :)

PR #515, opened 2022-01-23
[FIX] Switch ArrayCacheProvider to use Illuminate\Cache

### LINKED ISSUE #493: [BUG] Error Class "Doctrine\Common\Cache\ArrayCache" not found
### 1.7, 8.45

I configured a brand new Laravel application and installed the latest version of `Laravel Doctrine`  when trying to navigate to a page or execute an `artisan` command I get that error. Do you know why would this be happening?

dpslwk:
fix #493
I know `symfony/cache` was suggested in the issue but we already use `illuminate/cache` for most fo the other drivers, so its a quick change

we should be able to up merge this to 1.6 and 1.7 branch just resolving the composer  conflicts to bump "illuminate/cache": "^6.0", version as needed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,11 +17,11 @@
     ],
     "require": {
         "php": "^7.2|^8.0",
-        "doctrine/cache": "^1",
         "doctrine/dbal": "^2.13",
         "doctrine/orm": "^2.6",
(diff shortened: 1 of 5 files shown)


dpslwk:
nvm so not as easy as I thought

PR #523, opened 2022-05-09
[Fix] Allow symfony/serializer ^5.0 to be used

xaben:
- Add back to vendors symfony/serializer ^5.0 as a supported version
- Install symfony/yaml as was getting test failures without it?

I'm trying to upgrade our project to Laravel 9 and laravel-doctrine 1.8, but since we also use [googleads-php-lib](https://github.com/googleads/googleads-php-lib) and it does not list v6.0 as supported yet, I'm running into a dependency conflict.

I've opened an [issue](https://github.com/googleads/googleads-php-lib/issues/739) there as well, just want to get both libraries  side by side.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "illuminate/support": "^9.0",
         "illuminate/validation": "^9.0",
         "illuminate/view": "^9.0",
-        "symfony/serializer": "^6.0",
+        "symfony/serializer": "^5.0|^6.0",
         "symfony/yaml": "^5.0|^6.0"
     },
     "require-dev": {


eigan:
Any idea why symfony/serializer 5 and lower was dropped @dpslwk? Seems fine for me to change this back. Was changed as part of L9: 263ca2164ef5de7bbe1cc0f593e38c241f40c2e8 

Looks like it's only used in the Jsonable trait.

dpslwk:
when I was doing L9 I followed the framework's changes and just listed the latest version https://github.com/laravel/framework/commit/03fa6a2566626920738d2c27eec20aad75be4561
also most of the older versions are EOL
I think we would be on to add back `^5.0`, but we are not exactly check all these version in a CI matrix, that's only likely to test against `^6.0`

I'm no so sure about just adding in `symfony/yaml` just on a questionable failing test, could you remove that from this PR and see if the GH action actually fails with out it or provide the actual failed test details?

xaben:
Thanks for looking into this @dpslwk , I've removed the symfony/yaml requirement and now both actions have failed, I had the exact same errors when testing locally last time. I tested and I get same failures on the clean `1.8` branch currently.

As far as I understand `symfony/serializer^5.0` resolves to 5.4 which has same support cycle as the 5.4 LTS release of Symfony (End of bug fixes: November 2024 / End of security fixes: November 2025).

Regarding the CI matrix maybe it's worth tweaking to have a `composer update --prefer-lowest` flag as one of the variations? I can try and add it, if you are happy to merge such a change.

dpslwk:
ok so for yaml it looks like something in has changed since the ci was last ran, some dependancy that was include yaml is now no longer pulling it in, which is fine, we can add that back in

eigan:
@xaben Could you rebase and squash on top of 1.8 now? :)

xaben:
@eigan Rebased and squashed, tests are now passing :)

eigan:
Released with 1.8.1. Sorry for the delay!

PR #544, opened 2023-02-20
[2.0] Disable autodect of cache

eigan:
The new ORMSetup::createConfiguration() will automatically create cache configuration if you
do not specify cache as the third argument.
LaravelDoctrine supports setting different cache on each area, we do this in a later step.


dpslwk (changes requested):
looks good, just need to remove the now unused import

dpslwk (approved):

PR #545, opened 2023-02-21
Upgrade dependencies

eigan:
- Update to more recent versions. Its hard to keep support for previous dependencies.
- Add phpstan, ignore lots of issues because I do not have time to fix, but new issues should be reported.
-


dpslwk:
looks ok to me from what I understand, and defiantly support the reasoning

dpslwk (approved):

PR #556, opened 2023-06-07
[FEATURE] Add driver options to all drivers

keithbrink:
Continuing with #260, this provides support for setting driver options on all connection drivers.

This also aligns the setting with Laravel's setting, using "options" in the config rather than "driverOptions" ([link](https://github.com/laravel/framework/blob/2226e6baefd3f47ff307096b201546c46e855d11/src/Illuminate/Database/Connectors/Connector.php#L95)).


eigan:
I think the configuration here follows the keys from https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/configuration.html, not sure if should rename this setting, as Laravel is never involved anyway. Does the other keys match laravel settings?

keithbrink:
@eigan Yeah, the other keys are following Laravel's key names:

dbname -> database
user -> username

eigan:
Thanks!

PR #559, opened 2023-08-08
[FIX] selectCount() does not exists. maybe: select()

77web:
validation rules like `exists:App\Entities\Foo,1` does not work after https://github.com/laravel-doctrine/orm/commit/57612a2aefbda6572760a4e322b424b15eb072cc

- use select()


eigan:
Typo, thanks! 🙈

PR #561, opened 2023-08-28
[FIX]  primaryreplica connection init

sserbin:
this was accidentally passing the tests
(the configuration was still accepted even though it was just empty and the EM was setup with a non primary read connection but a regular one) which I've also made a fix for to verify the resulting connection is indeed a PrimaryReadReplicaConnection

- fix primary read replica connection giving error upon init


eigan:
Thanks!

PR #566, opened 2023-10-13
Changed docblock for getRepository() to satisfy PHPStan

soltmar:
[FIX] - Satisfies PHPStan's "Call to an undefined method Doctrine\Persistence\ObjectRepository::createQueryBuilder()."


eigan:
Thanks 👍🏻

PR #568, opened 2023-10-25
[FIX] Silence Passing null to class_exists() deprecation warning

soltmar:
Since PHP 8.1 passing null to class_exists() is deprecated. I have re-factored SubstituteBindings::getClassName() to silence deprecation warning

This change could possibly be ported to other branches as well.


eigan:
Thanks 👍🏻

PR #573, opened 2024-03-02
Laravel 11.x Compatibility

laravel-shift:
This is an automated pull request from [Shift](https://laravelshift.com) to update your package code and dependencies to be compatible with Laravel 11.x.

**Before merging**, you need to:

- Checkout the `l11-compatibility` branch
- Review **all** comments for additional changes
- Thoroughly test your package

If you do find an issue, please report it by commenting on this PR to help improve future automation.

PR #575, opened 2024-05-06
[FEATURE] Laravel 11 update

77web:
Updated DocrineUserProvider and versions for some dependencies from #573.
The updates for phpunit/phpunit and doctrine/orm should be done separately.
Updated DoctrineUserProvider and Authenticable in this reposiory for Laravel 11's newly added methods in Auth interfaces.

- Laravel11 update


src/DoctrineServiceProvider.php
```diff
@@ -88,6 +88,7 @@ protected function ensureValidatorIsUsable()
         if (!$this->isLumen()) {
             return;
         }
+        assert(property_exists($this->app, 'availableBindings'));
```
eigan, line 91: "added assert for lumen"

What does this mean?

77web: `$this->app` has "availableBindings" only when it is used in lumen.
Static analysis failed for `Illuminate\Contracts\Foundation\Application` does  not have "availableBindings" property:
`Access to an undefined property                                     Illuminate\Contracts\Foundation\Application::$availableBindings. `
https://github.com/laravel-doctrine/orm/actions/runs/8966821073/job/24623101202

The method `DoctrineServiceProvider::ensureValidatorIsUsable()` is for lumen(as shown in line 88), so the assert should be always correct here.

eigan: Allright, thanks 👍🏻

eigan:
Thanks @77web! Good job 🙂

bartjuh4:
@eigan when can we expect the new release?

eigan:
> @eigan when can we expect the new release?

Should be out. Do composer update. Just havent written a release here om github

PR #579, opened 2024-09-19
Add SSL certificate configuration keys to database connections

keithbrink:
The configuration keys in this PR are required for some database connections, but are not currently passed along from the Laravel configuration.

For reference: https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/configuration.html#mysqli


eigan:
Thanks!

PR #580, opened 2024-09-25
[FEATURE] Feature/orm3

TomHAnderson:
This PR adapts to ORM 3 and DBAL 4.  

This is ready to merge to 3.0.  Unit tests all passed, but a handful had to be skipped to allow for more time.


TomHAnderson:
All unit tests pass

src/Configuration/MetaData/MetaDataManager.php
```diff
@@ -12,7 +12,7 @@ class MetaDataManager extends Manager
      */
     public function getDefaultDriver()
     {
-        return 'annotations';
+        return 'xml';
```
eigan: Isnt the attributes driver a more natural default driver? Symfony uses this as default.

eigan (commented):
Good job. I think I  should create new target branch for this.

src/Configuration/MetaData/MetaDataManager.php
TomHAnderson: The most technically correct answer is `xml`.  However most users of Doctrine use the easy way out and use annotations, now attributes.  If the goal of this project is to provide the most simple possible Doctrine metadata configuration then `attributes` is the right answer.  Given the history of this project recommending annotations, I agree attributes would be the simple, easy, and best default.

TomHAnderson:
@eigan There are so many differences between my PR and 2.0 that I recommend a new branch, 3.0.x, be created.  I've got more work to do on this, too.  I intend to 

* Implement the Doctrine coding standard
* Move to psalm from phpstan

Since these two goals are separate from the goal of this PR, I request you create that new branch right away, knowing it won't pass actions checks right away.

TomHAnderson:
@eigan the base has been changed.  This PR is ready to merge to 3.0

composer.json
```diff
@@ -16,11 +16,11 @@
         }
     ],
     "require": {
-        "php": "^8.0",
-        "doctrine/annotations": "^2",
-        "doctrine/dbal": "^3.2",
-        "doctrine/orm": "^2.14",
-        "doctrine/persistence": "^3",
+        "php": "^8.1",
+        "doctrine/cache": "^2.2",
```
derrabus, line 20: Do you really still need Doctrine Cache?

TomHAnderson: Removing it, and the unit tests still pass.

TomHAnderson: This change will be in a future PR.  I'm waiting to see how much support this community will give my work on the library.

TomHAnderson:
I have two other PRs, each building on this one, that I'll queue up to review once this is merged.  What's your normal timeline to get refactoring work like this merged into a new branch?

TomHAnderson:
Moved to https://github.com/laravel-doctrine/orm/pull/582

PR #641, opened 2024-10-29
Installed testbench; removed problematic InteractsWithEntities trait and test

TomHAnderson:
This important update installs Testbench so the ServiceProvider can be tested.


k0ka:
Why remove InteractsWithEntites? It was quite useful helper class. And there is no alternative for it.

PR #644, opened 2024-11-01
Revert return to null if manager is not found; add bool to throw exce…

TomHAnderson:
…ption instead

### DIFF
--- a/src/IlluminateRegistry.php
+++ b/src/IlluminateRegistry.php
@@ -302,7 +302,7 @@ final class IlluminateRegistry implements ManagerRegistry
      *
      * @param class-string $className A persistent object class name.
      */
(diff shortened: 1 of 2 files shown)


src/IlluminateRegistry.php
```diff
@@ -302,7 +302,7 @@ public function getRepository(string $persistentObject, string|null $persistentM
      *
      * @param class-string $className A persistent object class name.
```
Hermaniandr, line 303: We could add `@return`  annotation, so static tools (e.g. `phpstan`) will understand that return type depends on argument

`* @return ($throwExceptionIfNotFound is true ? ObjectManager : ObjectManager|null)`

Hermaniandr (approved):

PR #647, opened 2024-11-27
fix: load entity factories

### LINKED ISSUE #645: Unable to locate factory with name [default] [App\Entities\User]
> If you have a question, and not a bug or proposal, post in the discussions.
> https://github.com/laravel-doctrine/orm/discussions 

# laravel-doctrine/orm version & Laravel version

    "require": {
        "php": "^8.2",
        "gedmo/doctrine-extensions": "^3.17",
        "laravel-doctrine/extensions": "^2.0",
        "laravel-doctrine/migrations": "^3.3",
        "laravel-doctrine/orm": "^3.0",
        "laravel/framework": "^11.9",
        "laravel/tinker": "^2.9"
    },
    "require-dev": {
        "fakerphp/faker": "^1.23",
        "laravel/pail": "^1.1",
        "laravel/pint": "^1.13",
        "laravel/sail": "^1.26",
        "mockery/mockery": "^1.6",
        "nunomaduro/collision": "^8.1",
        "phpunit/phpunit": "^11.0.1"
    },

# Expected behaviour

When running factories (either in tests or seeder), the factories in database/factories should be properly located.

# Actual behaviour

When running this example seeder:

```php
<?php

namespace Database\Seeders;

use App\Entities\User;
use Illuminate\Database\Seeder;

class DatabaseSeeder extends Seeder
{
    /**
     * Seed the application's database.
     */
    public function run(): void
    {
        entity(User::class)->create();
    }
}

``` 
with php artisan db:seed following error happens:

 InvalidArgumentException 

Unable to locate factory with name [default] [App\Entities\User]

at vendor/laravel-doctrine/orm/src/Testing/FactoryBuilder.php:205
  201▕      */
  202▕     protected function makeInstance(array $attributes = []): mixed
  203▕     {
  204▕         if (! isset($this->definitions[$this->class][$this->name])) {
➜ 205▕             throw new InvalidArgumentException('Unable to locate factory with name [' . $this->name . '] [' . $this->class . ']');
  206▕         }
  207▕ 
  208▕         $definition = call_user_func($this->definitions[$this->class][$this->name], $this->faker, $attributes);
  209▕ 

# Steps to reproduce 

1. Fresh install of L11 with packages 
  gedmo/doctrine-extensions
  laravel-doctrine/extensions
  laravel-doctrine/migrations
  laravel-doctrine/orm
  
2. Add entity, for example:

```php
<?php

namespace App\Entities;

use App\Traits\MustVerifyEmail;
use DateTime;
use DateTimeImmutable;
use Doctrine\DBAL\Types\Types;
use Doctrine\ORM\Mapping as ORM;
use Gedmo\Mapping\Annotation as Gedmo;
use Illuminate\Auth\Passwords\CanResetPassword;
use Illuminate\Contracts\Auth\Authenticatable;
use Illuminate\Contracts\Auth\CanResetPassword as CanResetPasswordContract;
use Illuminate\Contracts\Auth\MustVerifyEmail as MustVerifyEmailContract;
use LaravelDoctrine\ORM\Notifications\Notifiable;

#[ORM\Entity]
#[ORM\Table(name: "users")]
class User implements Authenticatable, CanResetPasswordContract, MustVerifyEmailContract
{
    use CanResetPassword, \LaravelDoctrine\ORM\Auth\Authenticatable, MustVerifyEmail, Notifiable;

    #[ORM\Id]
    #[ORM\GeneratedValue(strategy: "AUTO")]
    #[ORM\Column(type: 'integer')]
    protected int $id;

    #[ORM\Column(type: 'string', length: 255)]
    protected string $name;

    #[ORM\Column(type: 'string', length: 255, unique: true)]
    protected string $email;

    #[ORM\Column(type: 'datetime', nullable: true)]
    protected ?DateTime $emailVerifiedAt = null;

    #[ORM\Column(type: Types::DATETIME_IMMUTABLE)]
    #[Gedmo\Timestampable]
    protected ?DateTimeImmutable $createdAt = null;

    #[ORM\Column(type: Types::DATETIME_IMMUTABLE)]
    #[Gedmo\Timestampable]
    protected ?DateTimeImmutable $updatedAt = null;

    public function __construct() {}

    public function getId(): int
    {
        return $this->id;
    }

    public function getKey(): int
    {
        return $this->getId();
    }

    public function setId(int $id): void
    {
        $this->id = $id;
    }

    public function getName(): string
    {
        return $this->name;
    }

    public function setName(string $name): void
    {
        $this->name = $name;
    }

    public function getEmail(): string
    {
        return $this->email;
    }

    public function setEmail(string $email): void
    {
        $this->email = $email;
    }

    public function getEmailVerifiedAt(): ?DateTime
    {
        return $this->emailVerifiedAt;
    }

    public function setEmailVerifiedAt(?DateTime $emailVerifiedAt): void
    {
        $this->emailVerifiedAt = $emailVerifiedAt;
    }

    public function getCreatedAt(): DateTimeImmutable
    {
        return $this->createdAt;
    }

    public function getUpdatedAt(): DateTimeImmutable
    {
        return $this->updatedAt;
    }
}
``` 

3. Create factory to database/factories/UserFactory.php

```php
<?php

use Faker\Generator as Faker;
use Illuminate\Support\Facades\Hash;
use Illuminate\Support\Str;

$factory->define(App\Entities\User::class, function (Faker $faker, array $attributes) {
    return [
        'name' => $faker->name,
        'email' => $faker->unique()->safeEmail,
        'password' => Hash::make('password'),
        'emailVerifiedAt' => $attributes['emailVerifiedAt'] ?? new DateTime,
        'rememberToken' => Str::random(10),
    ];
});

``` 

4. Create seeder:

```php
<?php

namespace Database\Seeders;

use App\Entities\User;
use Illuminate\Database\Seeder;

class DatabaseSeeder extends Seeder
{
    /**
     * Seed the application's database.
     */
    public function run(): void
    {
        entity(User::class)->create();
    }
}

``` 
5. Run command > php artisan db:seed

The app works normally, no bugs or errors there but the factories are not currently located for some reason.

Thanks!

sstutz:
fixes #645 

I've never used testbench so I am not 100% sure whether the configuration changes are correct but as far as I could tell it wasn't using the ./workbench directory at all hence the new `laravel: ./workbench` directive.

With the change to use the correct workbench directory I had to add a handful of new files and directories in order to make the pipeline happy so they are not really part of the scope of the fix but I had to add them anyway.

### DIFF
--- a/src/DoctrineServiceProvider.php
+++ b/src/DoctrineServiceProvider.php
@@ -31,12 +31,15 @@ use LaravelDoctrine\ORM\Console\SchemaValidateCommand;
 use LaravelDoctrine\ORM\Exceptions\ExtensionNotFound;
 use LaravelDoctrine\ORM\Extensions\ExtensionManager;
 use LaravelDoctrine\ORM\Notifications\DoctrineChannel;
+use LaravelDoctrine\ORM\Testing\Factory as EntityFactory;
 use LaravelDoctrine\ORM\Validation\PresenceVerifierProvider;
 
 use function assert;
 use function class_exists;
 use function config;
 use function config_path;
+use function database_path;
+use function fake;
 
 class DoctrineServiceProvider extends ServiceProvider
 {
@@ -68,6 +71,7 @@ class DoctrineServiceProvider extends ServiceProvider
         $this->registerExtensions();
         $this->registerConsoleCommands();
         $this->registerCustomTypes();
+        $this->registerEntityFactory();
         $this->registerProxyAutoloader();
 
         if (! $this->shouldRegisterDoctrinePresenceValidator()) {
@@ -267,6 +271,20 @@ class DoctrineServiceProvider extends ServiceProvider
 
         $this->app->make(ChannelManager::class)->extend($channel, static function ($app) {
             return new DoctrineChannel($app['registry']);
+        });
+    }
+
+    /**
+     * Register the Entity factory instance in the container.
+     */
+    protected function registerEntityFactory(): void
+    {
+        $this->app->singleton(EntityFactory::class, static function ($app) {
+            return EntityFactory::construct(
+                fake(),
+                $app->make('registry'),
+                database_path('factories'),
+            );
         });
     }
 
--- a/workbench/database/factories/UserEntityFactory.php
+++ b/workbench/database/factories/UserEntityFactory.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Workbench\Database\Factories;
+
+use Faker\Generator;
+use LaravelDoctrine\ORM\Testing\Factory;
+use Workbench\App\Entities\User;
+
+/** @var Factory $factory */
+$factory->define(User::class, static function (Generator $faker, array $attributes = []) {
+    return [
+        'name' => $faker->name(),
+        'email' => $faker->safeEmail,
+        'password' => 'password',
+    ];
+});
+
+$factory->defineAs(User::class, 'test', static function (Generator $faker, array $attributes = []) {
+    return [
+        'name' => 'test',
+        'email' => 'test@test.tld',
+        'password' => 'password',
+    ];
+});
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ composer.lock
 /tests/Stubs/storage/doctrine.generated.php
 .idea
 laravel-doctrine-orm.iml
+/workbench/bootstrap/cache/*
+!/workbench/bootstrap/cache/.gitkeep
+/workbench/storage/logs/*
+/workbench/vendor
--- a/testbench.yaml
+++ b/testbench.yaml
@@ -1,12 +1,13 @@
 providers:
   - LaravelDoctrine\ORM\DoctrineServiceProvider
 
+laravel: ./workbench
 workbench:
   start: '/'
   install: true
   health: false
   discovers:
-    web: fale
+    web: false
     api: false
     commands: false
     components: false
--- a/tests/Feature/DoctrineServiceProviderEntityFactoryTest.php
+++ b/tests/Feature/DoctrineServiceProviderEntityFactoryTest.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaravelDoctrineTest\ORM\Feature;
+
+use Doctrine\ORM\Configuration;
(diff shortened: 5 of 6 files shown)


src/DoctrineServiceProvider.php
```diff
@@ -270,6 +275,24 @@ public function extendNotificationChannel(): void
         });
     }
 
+    /**
+     * Register the Entity factory instance in the container.
+     */
+    protected function registerEntityFactory(): void
+    {
+        $this->app->singleton(FakerGenerator::class, static function ($app) {
```
SanderSander: I got the same issue and saw your PR 

This shouldn't be needed, the Generator should be solveable. 
This also prevents moving the faker dependency from the `require-dev` section. 

I got the following temporarily in my `AppServiceProvider` 

```
$this->app->singleton(Factory::class, static function ($app) {
            return Factory::construct(
                $app->make(Generator::class),
                $app->make('registry'),
                database_path('factories'),
);
});
```

This seems to do work fine, so no need to register the `Generator` into the container

TomHAnderson: I agree about the require-dev section home for faker.  Can this be solved using an instantiator?  

Gonna need to not move faker from require-dev and put some unit tests in too.

sstutz:
@TomHAnderson can u take a look again?

workbench/database/factories/UserEntityFactory.php
```diff
@@ -0,0 +1,24 @@
...
+use Faker\Generator;
+use LaravelDoctrine\ORM\Testing\Factory;
+use Workbench\App\Entities\User;
+
+/** @var Factory $factory */
+$factory->define(User::class, static function (Generator $faker, array $attributes = []) {
+    return [
+        'name' => $faker->name(),
+        'email' => $faker->safeEmail,
+        'password' => 'password',
```
TomHAnderson, line 14: If password is always set to 'password' how can the unit test pass looking for 'abc'?

sstutz: these are just default values which can be overwritten by passing in an array to make() which is what I do in the test file on line 39:
`$user = entity(User::class)->make(['password' => 'abc']);`

these values are merged with the default values when hydrating the object, see https://github.com/laravel-doctrine/orm/blob/04915c1ecd9e888e166c1389d07605527a45e991/src/Testing/FactoryBuilder.php#L230

PR #654, opened 2025-02-02
rememberToken should allow nullable

bartjuh4:
Return type for getRememberToken method should also support nullable


TomHAnderson:
Please run `composer test` before pushing commits.

This PR is closed in favor of #655

PR #662, opened 2025-03-19
Fixed deprecation notice for php 8.4

websmurf:
We are not ready for the upgrade to ORM 3.x yet, so this is a small fix for deprecations on php 8.4

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -436,12 +436,12 @@ parameters:
 			path: src/Console/DumpDatabaseCommand.php
 
 		-
-			message: "#^Parameter \\#1 \\$em of method LaravelDoctrine\\\\ORM\\\\Console\\\\DumpDatabaseCommand\\:\\:dump\\(\\) expects string, array\\|bool\\|string\\|null given\\.$#"
+			message: "#^Parameter \\#1 \\$em of method LaravelDoctrine\\\\ORM\\\\Console\\\\DumpDatabaseCommand\\:\\:dump\\(\\) expects string, array\\|string\\|true given\\.$#"
 			count: 1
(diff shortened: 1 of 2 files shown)


websmurf:
I've had a look at the errors in phpstan, but they don't seem to be related to my change as far as I can tell.
What the correct procedure for this?

TomHAnderson:
You're right, they aren't related to your change.  This is probably due to a fluid phpstan version.  I'll look into them.

TomHAnderson:
This is the last CI run on the 2.0 branch:
https://github.com/laravel-doctrine/orm/actions/runs/11431449791
The logs have expired.  

There are a couple things that could be done:
* Fix the PHPStan issues
* Remove PHPStan from this archaic 2.0 branch

I'm OK with either fix.  But it must be noted that I came on to this project to build v3.  I didn't take on the responsibility of the 2.0 branch.  That being said, I'm very willing to work with contributors on any contributions to the 2.0 branch.

It would be helpful if anyone from the 2.0 team could chime in here with their view.

websmurf:
@TomHAnderson the errors from Console/DumpDatabaseCommand.php are currently in the baseline as well, so that's an easy fix. I'll just regenerate the baseline with the new phpstan error format :-)

The deprecation of Doctrine\Common\Proxy\Autoloader is known as well, and will be fixed in the 3.0 upgrade.
The convertToLaravelPaginator will have implications for users of laravel doctrine as well I think because it most likely will force users to hint them as well.

TomHAnderson:
Fixed at https://github.com/laravel-doctrine/orm/blob/3.3.x/src/Extensions/Extension.php#L12

SPie:
@TomHAnderson Can I ask why this MR was closed? This was a fix for the 2.0 branch.

Issue #663, opened 2025-05-21
EntityManager is resolved during package discovery


LegendEffects:
Hello, when composer dumps it's autoload Laravel automatically runs the `package:discover` command, when doing this it will try to initialise a connection if you have any custom types defined (related: https://github.com/laravel-doctrine/orm/issues/648) and cause an error if you don't have a configured database.

This can be avoided by moving the EntityManagerProvider dependency outside of the constructor thus avoiding any attempt to start Doctrine when we don't need it yet. I tested this just by commenting out registering commands in the ServiceProvider and it was able to do package discovery without errors.

I see why this is being done so that doctrine commands can just be directly extended, there's got to be some lazy way of doing this.

TomHAnderson:
Should be fixed here: 
https://github.com/laravel-doctrine/orm/releases/tag/3.1.1

LegendEffects:
Fixed for me, thanks

PR #664, opened 2025-05-21
Lazily resolve ManagerRegistry in EntityManagerProvider

LegendEffects:
Fixes an issue where a connection will always be attempted when running any command (including package discovery) if there are custom types installed.

https://github.com/laravel-doctrine/orm/issues/663

### DIFF
--- a/src/Console/EntityManagerProvider.php
+++ b/src/Console/EntityManagerProvider.php
@@ -7,18 +7,30 @@ namespace LaravelDoctrine\ORM\Console;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Tools\Console\EntityManagerProvider as DoctrineEntityManagerProvider;
 use Doctrine\Persistence\ManagerRegistry;
+use Illuminate\Contracts\Container\Container;
 
 use function assert;
 
 class EntityManagerProvider implements DoctrineEntityManagerProvider
 {
-    public function __construct(private ManagerRegistry $managerRegistry)
+    private ManagerRegistry|null $managerRegistry = null;
+
+    public function __construct(private Container $container)
     {
+    }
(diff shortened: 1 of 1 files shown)


src/Console/EntityManagerProvider.php
```diff
@@ -7,18 +7,30 @@
...
+            $this->managerRegistry = $this->container->make(ManagerRegistry::class);
+        }
+
+        return $this->managerRegistry;
+    }
+
     public function getDefaultManager(): EntityManagerInterface
     {
-        $entityManager = $this->managerRegistry->getManager();
+        $entityManager = $this->getManagerRegistry()->getManager();
```
R3VoLuT1OneR: @LegendEffects 
Following the doctrine filosophy you shouldn't cache the get manager.
Just do the `app(ManagerRegistry::class)` or `$this->container->make(ManagerRegistry::class)` where you need it.

The `make` not gonna create another object, it's defined as singleton it's gonna inject previously created manager registry.

Why?
Because you gonna have control over the injected Manager Registry object.
You want to have this control for tests.

LegendEffects: Good point, I was just keeping the same code flow as before, I avoided using the `app` helper as there was no other instances of it being used in the package

TomHAnderson:
I've updated the test runner in #665 
Please merge or rebase this PR with 3.1.x

TomHAnderson:
Tests fail because the new code in this patch isn't covered

LegendEffects:
There are no tests present for the entire Console namespace, adding tests to this would also pretty much just be testing the PSR container so feels pointless.

TomHAnderson:
Thanks for the feedback.  I agree.

PR #670, opened 2025-08-20
feat: use overrides for connection settings

keithbrink:
We recently added the `application_name` setting to our database configuration, which is supported by Laravel but was not passed into Doctrine by this package.

I'm proposing that instead of specifying the exact set of keys passed to Doctrine, the package instead overrides the required ones and passes along the remaining array as is.


keithbrink:
I believe the test failures here are unrelated to these changes and require the test suite to be fixed on 3.1.x.

TomHAnderson:
Merged in #684

PR #676, opened 2025-11-24
fix test suite by adding expectation for native lazy objects option

keithbrink:
Doctrine ORM 3.4 adding an option in the configuration for native lazy objects. That broke the test suite since the configuration is being mocked. This adds an expectation for that configuration option check to fix the test suite.


TomHAnderson:
Fixed in #683

PR #690, opened 2026-02-08
pillmayer-daniel:patch-1

TomHAnderson:

PR #693, opened 2026-03-04
Fix performance issue in getManagerForClass

### LINKED ISSUE #692: Performance issue in DoctrineManagerRegistry::getManagerForClass()
`DoctrineManagerRegistry::getManagerForClass()` is running `$entityManager->getMetadataFactory()->getAllMetadata()` this is a HEAVY operation that should not be executed in this context.

This was previously fixed in https://github.com/laravel-doctrine/orm/pull/288 , but later reverted in https://github.com/laravel-doctrine/orm/pull/628

Checking if the class is not transient should be enough, like this:
```php
    public function getManagerForClass(string $class): ObjectManager|null
    {
        $proxyClass = new \ReflectionClass($class);
        if ($proxyClass->isAnonymous()) {
            return null;
        }

        if ($proxyClass->implementsInterface(Proxy::class)) {
            $parentClass = $proxyClass->getParentClass();

            if ($parentClass === false) {
                return null;
            }

            $class = $parentClass->getName();
        }

        foreach ($this->getManagers() as $manager) {
            if (!$manager->getMetadataFactory()->isTransient($class)) {
                return $manager;
            }
        }

        return null;
    }
```

This is what's done in symfony applications ( https://github.com/doctrine/persistence/blob/4.1.x/src/AbstractManagerRegistry.php#L128 )

tomme87:
Avoided using `getAllMetadata()`, since that is a heavy operation. Checking if the class is not transient is enough.

Fixes #692

### DIFF
--- a/src/IlluminateRegistry.php
+++ b/src/IlluminateRegistry.php
@@ -324,16 +324,8 @@ final class IlluminateRegistry implements ManagerRegistry
         }
 
         foreach ($this->getManagers() as $entityManager) {
-            if ($entityManager->getMetadataFactory()->isTransient($className)) {
-                // @codeCoverageIgnoreStart
-                continue;
-                // @codeCoverageIgnoreEnd
-            }
-
-            foreach ($entityManager->getMetadataFactory()->getAllMetadata() as $metadata) {
-                if ($metadata->getName() === $className) {
-                    return $entityManager;
-                }
+            if (!$entityManager->getMetadataFactory()->isTransient($className)) {
+                return $entityManager;
             }
         }
 
--- a/src/IlluminateRegistry.php
+++ b/src/IlluminateRegistry.php
@@ -324,7 +324,7 @@ final class IlluminateRegistry implements ManagerRegistry
         }
 
         foreach ($this->getManagers() as $entityManager) {
-            if (!$entityManager->getMetadataFactory()->isTransient($className)) {
+            if (! $entityManager->getMetadataFactory()->isTransient($className)) {
                 return $entityManager;
             }
         }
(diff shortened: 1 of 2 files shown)


tomme87:
I see now that https://github.com/laravel-doctrine/orm/pull/690 fixes the same issue. But with another solution.  I don't think the extra check in the other PR is necessary though.

TomHAnderson:
I've held off on #690 because I wasn't sure of side effects.  What do you think of this?

### Comparison: `isTransient()` vs. `getAllClassNames()` loop

In the context of resolving the performance issues in `getManagerForClass`, here is how the two approaches stack up:

| Feature | `$isTransient($className)` | `foreach` with `getAllClassNames()` |
| :--- | :--- | :--- |
| **Performance** | **High.** Performs a direct lookup in the metadata factory. | **Lower.** Requires an $O(n)$ string comparison loop through all classes. |
| **Side Effects** | **High Risk.** Can trigger full metadata loading, which is a "heavy operation" that may cause logic loops. | **Safe.** Accesses the driver's list of names without forcing Doctrine to fully "boot" the metadata. |
| **Stability** | Faster, but prone to "Metadata not found" exceptions if the class isn't mapped. | More robust for checking existence across multiple managers. |

---

### Analysis
* **The First Line (`isTransient`):** As noted by [tomme87](https://github.com/tomme87) in [Issue #692](https://github.com/laravel-doctrine/orm/issues/692), this is a major optimization over the original `getAllMetadata()` call because it avoids loading the entire schema just to check a single class.
* **The Second Line (The Loop):** This was the alternative proposed in [PR #690](https://github.com/laravel-doctrine/orm/pull/690). While safer in complex multi-manager setups, it introduces a performance hit by iterating through every registered entity name.

**The Verdict:** For this specific fix, **`isTransient()`** is the better balance. It solves the "heavy operation" problem of `getAllMetadata()` while remaining significantly more efficient than a manual loop.

tomme87:
I am no expert in doctrine and I have very rarely used a multi-manager setup. 
But I know that Symfony uses doctrines implementation of `getManagerForClass()`  (https://github.com/doctrine/persistence/blob/4.1.x/src/AbstractManagerRegistry.php#L128), which uses `isTransient()` without any extra check. And the few times I have actually used a multi-manager setup (in a symfony application), this has worked fine.

Given this and your summary, I agree that using `isTransient()` is better.

